PR #1, opened 2022-03-29
Adds cli functionality

Nino-K:


pkg/commands/start.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func Start(address string, udpLocalPort, tcpLocalPort int, IPv6 bool, hosts map[string]string) error {
+	//TODO: seperate the handler from start
+	srv, err := dns.Start(address, udpLocalPort, tcpLocalPort, IPv6, hosts)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Started srv %+v", srv)
```
mook-as, line 31: Shouldn't there be a `defer srv.Shutdown()` somewhere?

mook-as (changes requested):
Why do we have a subcommand here at all? Do we have plans for other commands? (If yes, where?)

pkg/commands/start.go
Nino-K: 👍  I will be adding more enhancements in the upcoming PRs, this PR us just adding the cmd stuff. There are still some changes that I will need to make in `pkg/dns`

Nino-K:
> Why do we have a subcommand here at all? Do we have plans for other commands? (If yes, where?)

The `command/start.go` tends to be a wrapper around the `pkg/dns` thing. Maybe we will be adding other verbs in the future e.g `host-resolver flush`, `reset` etc, maybe not 🤷 . But for now, I needed a place to add other stuff, since we will be pushing all our changes for `pkg/dns` to the upstream and we will not have that forked locally. For now, it follows the `PROCESS VERB` logic, however, I'm open to suggestions for renaming `commands` to something that makes more sense.

cmd/root.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright © 2022 SUSE LLC
+1
```
jandubois: What is this `1` here?

Also, do we really need this big license blob in every file? I don't see it in the Rancher Desktop Typescript sources, so why is this needed in the Go files?

Nino-K: Not sure how did that `1` got in there, probably from where I copied it from(now I need to figure out where I copied it from 😂 ). General practice for open-source software is to include the header license in all the files. We should probably make that decision to either include everywhere or what is the plan. Because I have seen the random appearance of license in file headers in `rancher-desktop`.

mook-as: Yes, we're _supposed_ to have those for Rancher Desktop TypeScript (and everything else) files too; we just haven't been doing it properly…

mook-as (approved):
LGTM.

We should start considering having tests.

Nino-K:
> LGTM.
> 
> We should start considering having tests.

For sure, I'm adding them as we speak. Next PRs will have them.

PR #3, opened 2022-03-31
adds tests

Nino-K:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,4 +17,5 @@ RUN go mod download
 COPY . ./
 
 RUN go build -o /host-resolver
-CMD /host-resolver run -a 127.0.0.1 -t 54 -u 53 -c "host.rd.internal=111.111.111.111,host2.rd.internal=222.222.222.222" & go test -v ./...
+
+CMD go test -v ./...
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.17-alpine
+# TODO: Investigate why Alpine 1.18 is broken :|
+
+WORKDIR /app
+
+COPY go.mod .
+COPY go.sum .
+
+ENV GO111MODULE=on
+# need CGO disabled since container has no cgo installed
+ENV CGO_ENABLED=0
+#see which DNS resolver is being used
+#ENV GODEBUG=netdns=go+2
+
+RUN go mod download
+
+COPY . ./
+
+RUN go build -o /host-resolver
+CMD /host-resolver run -a 127.0.0.1 -t 54 -u 53 -c "host.rd.internal=111.111.111.111,host2.rd.internal=222.222.222.222" & go test -v ./...
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-FROM golang:1.17-alpine
-# TODO: Investigate why Alpine 1.18 is broken :|
+FROM golang:1.18.0-stretch
 
 WORKDIR /app
 
@@ -7,15 +6,16 @@ COPY go.mod .
 COPY go.sum .
 
 ENV GO111MODULE=on
-# need CGO disabled since container has no cgo installed
-ENV CGO_ENABLED=0
+ENV CGO_ENABLED=1
 #see which DNS resolver is being used
 #ENV GODEBUG=netdns=go+2
 
+RUN apt update
+RUN apt install net-tools
 RUN go mod download
 
 COPY . ./
 
-RUN go build -o /host-resolver
+RUN go build -o /app/host-resolver
 
 CMD go test -v ./...
--- a/pkg/commands/start_test.go
+++ b/pkg/commands/start_test.go
@@ -1,0 +1,76 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package commands
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStart(t *testing.T) {
+	netstat(t)
+
+	t.Log("Checking for TCP port on 54")
+	tcpPort := "54"
+	_, err := net.Listen("tcp", fmt.Sprintf(":%s", tcpPort))
+	require.Errorf(t, err, "host-resolver is not listening on TCP port %s", tcpPort)
+
+	t.Log("Checking for UDP port on 53")
+	udpPort := "53"
+	_, err = net.Listen("udp", fmt.Sprintf(":%s", udpPort))
+	require.Errorf(t, err, "host-resolver is not listening on UDP port %s", udpPort)
+}
+
+func TestQueryTCP(t *testing.T) {
+	t.Log("Checking for TCP port on 54")
+	addrs, err := dnsLookup(t, "54", "tcp", "host.rd.internal")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
+	require.Exactly(t, addrs[0].String(), "111.111.111.111")
+}
+
+func TestQueryUDP(t *testing.T) {
+	t.Log("Checking for UDP port on 53")
+	addrs, err := dnsLookup(t, "53", "udp", "host2.rd.internal")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
+	require.Exactly(t, addrs[0].String(), "222.222.222.222")
+}
+
+func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]net.IP, error) {
+	resolver := net.Resolver{
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			dialer := net.Dialer{}
+			return dialer.DialContext(ctx, resolverProtocol, fmt.Sprintf(":%s", resolverPort))
+		},
+	}
+	t.Logf("[DNS] lookup on :%s and %s -> %s", resolverPort, resolverProtocol, domain)
+	// 10s timeout should be adequate
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	return resolver.LookupIP(ctx, "ip4", domain)
+
+}
+
+func netstat(t *testing.T) {
+	out, err := exec.Command("netstat", "-nlp").Output()
+	require.NoError(t, err, "netstat -nlp")
+	t.Logf("%s\n", out)
+}
--- a/pkg/commands/start_test.go
+++ b/pkg/commands/start_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"net"
 	"os/exec"
-	"strings"
 	"testing"
 	"time"
 
@@ -26,8 +25,8 @@ import (
 )
 
 var (
-	tcpPort = "54"
-	udpPort = "53"
+	tcpPort = "5353"
+	udpPort = "5353"
 )
 
 func TestStart(t *testing.T) {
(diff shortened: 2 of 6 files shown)


pkg/commands/start_test.go
```diff
@@ -0,0 +1,76 @@
...
+func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]net.IP, error) {
+	resolver := net.Resolver{
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			dialer := net.Dialer{}
+			return dialer.DialContext(ctx, resolverProtocol, fmt.Sprintf(":%s", resolverPort))
+		},
+	}
+	t.Logf("[DNS] lookup on :%s and %s -> %s", resolverPort, resolverProtocol, domain)
+	// 10s timeout should be adequate
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
Nino-K, line 104: Calling with timeout here because I want to capture some of the known issues with host-resolver timeouts. Otherwise, if you feel like this is unnecessary we can just call with `context.Background()`.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.17-alpine
+# TODO: Investigate why Alpine 1.18 is broken :|
+
+WORKDIR /app
+
+COPY go.mod .
+COPY go.sum .
+
+ENV GO111MODULE=on
+# need CGO disabled since container has no cgo installed
```
Nino-K: I meant the container has no `gcc` installed.

jandubois:
The changes to `pkg/dns.go` are unrelated to adding the tests, so should be in a separate PR (or at least separate commit).

jandubois:
I think the test is fine for now, but using dns at `127.0.0.1` and not enabling CGO means you will not be testing any real lookups to upstream nameservers.

Nino-K:
> 



> I think the test is fine for now, but using dns at `127.0.0.1` and not enabling CGO means you will not be testing any real lookups to upstream nameservers.

I agree, for now, I just wanted to have some scaffolding that we can build on top of. Ultimately I will enhance the tests to do a real life name resolutions. I just need to figure out why golang1.18 alpine image is so broken.

pkg/commands/start_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestStart(t *testing.T) {
+	netstat(t)
+
+	t.Log("Checking for TCP port on 54")
+	tcpPort := "54"
+	_, err := net.Listen("tcp", fmt.Sprintf(":%s", tcpPort))
+	require.Errorf(t, err, "host-resolver is not listening on TCP port %s", tcpPort)
```
mook-as: If the test fails (there was no error), don't we need to close the listener?

mook-as, line 41: Shouldn't we verify that the error was because the port is already in use (and not because, for example, because it's a privileged port)?

Actually, shouldn't we verify that it's because the port is in use _by the host resolver_?

pkg/commands/start_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"net"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStart(t *testing.T) {
```
mook-as, line 32: For testing, I think it would be better if the test itself launched the host resolver (rather than relying on the Dockerfile to do it). Among other things:
- It would make it easier to run locally
- We can centralize where we pick the port numbers (so it's not hard-coded in both the Dockerfile and here)

pkg/commands/start_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	t.Log("Checking for UDP port on 53")
+	udpPort := "53"
+	_, err = net.Listen("udp", fmt.Sprintf(":%s", udpPort))
+	require.Errorf(t, err, "host-resolver is not listening on UDP port %s", udpPort)
+}
+
+func TestQueryTCP(t *testing.T) {
+	t.Log("Checking for TCP port on 54")
+	addrs, err := dnsLookup(t, "54", "tcp", "host.rd.internal")
```
mook-as: We should use the [dedicated `.test` TLD](https://datatracker.ietf.org/doc/html/rfc2606#section-2) for tests (i.e. `host.rd.test` or whatever). My initial review comment was "we actually use `host.rancher-desktop.internal`" before realizing this  isn't using that…

pkg/commands/start_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Logf("[DNS] lookup on :%s and %s -> %s", resolverPort, resolverProtocol, domain)
+	// 10s timeout should be adequate
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	return resolver.LookupIP(ctx, "ip4", domain)
+
+}
+
+func netstat(t *testing.T) {
+	out, err := exec.Command("netstat", "-nlp").Output()
```
mook-as, line 110: This is Linux-only then? Makes sense given the Dockerfile, I guess… should mark it somewhere though (probably via a golang build tag).

mook-as (changes requested):
Shouldn't there be a `.github/workflows/*` to trigger the test?

pkg/commands/start_test.go
Nino-K: yeah, It would be a good idea.

pkg/commands/start_test.go
Nino-K: I agree, this is the approach I would like to take. I would like to be able to start and stop the process so we can test things like IPv6, etc. I was hoping to add that in the upcoming PRs or do you prefer it done here?

pkg/commands/start_test.go
Nino-K: Yes, I will add the build tags once I have the platform-specific stuff, didn't add them yet since it really screws up the `gopls` stuff, I was hoping to enjoy the lack of build tags for a few more days 😂

Nino-K:
> Shouldn't there be a `.github/workflows/*` to trigger the test?

👍  I will have to add the github runner stuff. I will put it in a separate PR.

pkg/commands/start_test.go
Nino-K: This is not checking the privileged port at all, it will check on any port that `host-resolver` starts. But right now it's is hard coded to `5x`.

pkg/commands/start_test.go
mook-as: I think IPv6 etc. is definitely out of scope for this PR.

I think making the test itself start/stop the process would be in scope though; the Dockerfile isn't really useful for testing (especially if we pick an unprivileged port).  Ideally we can just start host resolver and have it pick a random port and report what it was… but that might be too much work for this test.

jandubois: > Ideally we can just start host resolver and have it pick a random port and report what it was…

That was the plan outlined in #1899. But that then requires us do to traffic filtering to prevent external access to the service. So the updated plan (not documented anywhere yet) is to create an internal interface that is routable only on the host and from the VM, and then use the regular port (53) to listen. Also saves us from having to use `iptables` inside the VM to redirect the random ports back to port 53 on an internal interface (because the port for DNS is not configurable).

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.18.0-stretch
+
+WORKDIR /app
+
+COPY go.mod .
+COPY go.sum .
+
+ENV GO111MODULE=on
+# need CGO disabled since container has no cgo installed
+ENV CGO_ENABLED=1
```
mook-as, line 9: Wait, your comment says to _disable_ `CGO`, but here it's _enabled_?

pkg/commands/start_test.go
```diff
@@ -0,0 +1,108 @@
...
+	udpListener, err := net.Listen("udp", fmt.Sprintf(":%s", udpPort))
+	if udpListener != nil {
+		defer udpListener.Close()
+	}
+	require.Errorf(t, err, "host-resolver is not listening on UDP port %s", udpPort)
+
+	output := netstat(t)
+
+	exist := strings.Contains(string(output), fmt.Sprintf("%v/host-resolver", cmd.Process.Pid))
+	require.Exactly(t, exist, true, "Expected the same Pid")
```
mook-as: Use [`require.Contains`](https://pkg.go.dev/github.com/stretchr/testify/require#Contains) instead so that if there's an error we would get the actual output.

pkg/commands/start_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestQueryTCP(t *testing.T) {
+	t.Logf("Checking for TCP port on %s", tcpPort)
+	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+	defer cmd.Process.Kill()
+
+	addrs, err := dnsLookup(t, tcpPort, "tcp", "host.rd.test")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
+	require.Exactly(t, addrs[0].String(), "111.111.111.111")
```
mook-as: Maybe use [`requires.ElementsMatch`](https://pkg.go.dev/github.com/stretchr/testify/require#ElementsMatch) (or just [`Equal`](https://pkg.go.dev/github.com/stretchr/testify/require#Equal)) so that if there's an incorrect number of addresses we can see what they were.

pkg/commands/start_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestQueryUDP(t *testing.T) {
+	t.Logf("Checking for UDP port on %s", udpPort)
+	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+	defer cmd.Process.Kill()
+
+	addrs, err := dnsLookup(t, udpPort, "udp", "host2.rd.test")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
+	require.Exactly(t, addrs[0].String(), "222.222.222.222")
```
mook-as: Same as above — feed the actual list to the assertion so that we get better output.

pkg/commands/start_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func run(t *testing.T, ip, tcpPort, udpPort, hosts string) *exec.Cmd {
+	cmd := exec.Command("/host-resolver", "run", "-a", ip, "-t", tcpPort, "-u", udpPort, "-c", hosts, "&")
```
mook-as: Is the executable always at `/`? Shouldn't this be something relative to the current directory? (You may need to use `go run` instead or something…)

mook-as: There should not be a `&` at the end of the argument list; you're not spawning this through a shell, so `&` doesn't background things (it just gets passed to the application as an extra argument).

pkg/commands/start_test.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var (
+	tcpPort = "5353"
+	udpPort = "5353"
+)
+
+func TestStart(t *testing.T) {
+	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+	defer cmd.Process.Kill()
```
mook-as, line 34: Should we use `.Signal(SIGTERM)` instead, to gracefully exit? This sends a `SIGKILL` instead.

mook-as (changes requested):

pkg/commands/start_test.go
Nino-K: I would assume in the test environment `SIGKILL` would be more ideal to reduce the flakes since it can not be ignored or blocked. Also, there is a potential that `SIGTERM` may not kill the child processes. 🤷

pkg/commands/start_test.go
Nino-K: > Is the executable always at `/`? Shouldn't this be something relative to the current directory? (You may need to use `go run` instead or something…)

The building takes place in docker file, sure, I can move things back into the `/APPDIR./

Dockerfile
Nino-K: Yeah, not relevant anymore. I will remove the comment.

pkg/commands/start_test.go
Nino-K: I remember why I used `exactly` now. The `Equal` will not work because the slices will not be equal since net package creates/parses IPv4 slice with a length of IPv6 (` p := make(IP, IPv6len)`), not too sure why?! 

Also, the `requires.ElementMatch` seems to have a strange behavior that I didn't want to investigate too much time to figure out. `Error:      	element 111.111.111.111 appears more times in [111.111.111.111] than in [111.111.111.111]` this could be potentially extra bytes that is caused by the net.ParseIP

```
expected: []net.IP{net.IP{0x6f, 0x6f, 0x6f, 0x6f}}
actual  : []net.IP{net.IP{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xff, 0xff, 0x6f, 0x6f, 0x6f, 0x6f}}
```

I will investigate see what is going on now.

Nino-K: I guess it's a known [issue](https://stackoverflow.com/a/42921422), and both `ElementMatch` and `Equal` perform reflect.DeepEqual here: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L65

pkg/commands/start_test.go
jandubois: 👍 on using `SIGKILL` for the test.

pkg/commands/start_test.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func TestQueryTCP(t *testing.T) {
+	t.Logf("Checking for TCP port on %s", tcpPort)
+	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+	defer cmd.Process.Kill()
+
+	addrs, err := dnsLookup(t, tcpPort, "tcp", "host.rd.test")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
```
mook-as: Given that you managed to use `ElementsMatch`, please _don't_ also first check the length. The way you're checking the length means that if `len(addrs) == 2` the output is just "1 != 2", and _not_ what elements you have.

(FWIW, using at least [`require.Len`](https://pkg.go.dev/github.com/stretchr/testify/require#Len) would be better — but even better to just check that the two lists are equal.)

pkg/commands/start_test.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func TestQueryUDP(t *testing.T) {
+	t.Logf("Checking for UDP port on %s", udpPort)
+	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+	defer cmd.Process.Kill()
+
+	addrs, err := dnsLookup(t, udpPort, "udp", "host2.rd.test")
+	require.NoError(t, err, "Lookup IP failed")
+	require.Exactly(t, len(addrs), 1, "Expect only one address")
```
mook-as: Same as before — please ~never assert equality between a number and an array length.

mook-as (changes requested):

pkg/commands/start_test.go
Nino-K: > means that if len(addrs) == 2 the output is just "1 != 2", and not what elements you have.

Although, that is the criteria. For the addrs to only have 1 address and that is the fail condition but, agreed, now that the slices are being compared it seems not necessary.

mook-as (approved):

PR #6, opened 2022-04-04
Add upstream servers

Nino-K:


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.18.0-stretch
+
+WORKDIR /app
+
+COPY go.mod .
+COPY go.sum .
+
+ENV GO111MODULE=on
+# need CGO disabled since container has no cgo installed
```
Nino-K: No longer relevant

Nino-K:
@mook-as & @jandubois PTAL

pkg/commands/start.go
```diff
@@ -22,9 +22,16 @@ import (
...
+func Start(address string, udpLocalPort, tcpLocalPort int, IPv6 bool, hosts map[string]string, upstreamServers []string) error {
+	options := dns.ServerOptions{
+		Address:         address,
+		UDPPort:         udpLocalPort,
+		TCPPort:         tcpLocalPort,
+		IPv6:            IPv6,
+		StaticHosts:     hosts,
+		UpstreamServers: upstreamServers,
+	}
+	srv, err := dns.Start(options)
```
mook-as: 👍 passing in an options struct is better than a random list of parameters.

Would `dns.Start(dns.ServerOptions{…})` be easier?

pkg/dns/dns.go
```diff
@@ -45,25 +56,33 @@ func (s *Server) Shutdown() {
...
-			return nil, err
+			logrus.WithError(err).Warnf("failed to detect system DNS, falling back to %v", defaultFallbackIPs)
+			cc, err = newStaticClientConfig(defaultFallbackIPs)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
+
+	cc, err = newStaticClientConfig(upstreamServers)
```
mook-as: Shouldn't this be in an `else` (that is, don't use `upstreamServers` if it's empty)? Otherwise you're clobbering `cc`.

mook-as: Should the tests have caught this?

pkg/commands/start_test.go
```diff
@@ -106,6 +123,10 @@ func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]n
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
 
+func randomPort() string {
+	return fmt.Sprint(rand.Intn(65535-54) + 54)
```
mook-as: The minimum random port should be 1024 (but ideally > 32767)

pkg/commands/start_test.go
```diff
@@ -16,67 +16,84 @@ package commands
...
+	require.True(t, len(addrs) > 0, true, "Expect at least an address")
 }
 
-func run(t *testing.T, ip, tcpPort, udpPort, hosts string) *exec.Cmd {
-	cmd := exec.Command("/app/host-resolver", "run", "-a", ip, "-t", tcpPort, "-u", udpPort, "-c", hosts)
+func runHostResovler(t *testing.T, args []string) *exec.Cmd {
+	// add run command to the tip
+	args = append([]string{"run"}, args...)
+	// add background process to the tail
+	args = append(args, "&")
```
mook-as: What background process? You're not running the command with `/bin/sh`, there's nothing here to interpret that.

Also, you're using `cmd.Start()` (instead of `cmd.Run()`), it's never blocking anything in the first place…

pkg/commands/start_test.go
```diff
@@ -16,67 +16,84 @@ package commands
...
 	output := netstat(t)
 	require.Contains(t, string(output), fmt.Sprintf("%v/host-resolver", cmd.Process.Pid), "Expected the same Pid")
 }
 
-func TestQueryTCP(t *testing.T) {
-	t.Logf("Checking for TCP port on %s", tcpPort)
-	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+func TestQueryStaticHosts(t *testing.T) {
+	port := randomPort()
+	cmd := runHostResovler(t, []string{"-a", "127.0.0.1", "-t", port, "-u", port, "-c", "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222"})
```
mook-as: It would be good to use _different_ ports for TCP & UDP, to ensure that we're not accidentally using the other.

pkg/commands/start_test.go
```diff
@@ -16,67 +16,84 @@ package commands
...
-	t.Logf("Checking for UDP port on %s", udpPort)
-	cmd := run(t, "127.0.0.1", tcpPort, udpPort, "host.rd.test=111.111.111.111,host2.rd.test=222.222.222.222")
+func TestQueryUpstreamServer(t *testing.T) {
+	port := randomPort()
+	cmd := runHostResovler(t, []string{"-a", "127.0.0.1", "-t", port, "-u", port, "-s", "[8.8.8.8]"})
 	defer cmd.Process.Kill()
 
-	addrs, err := dnsLookup(t, udpPort, "udp", "host2.rd.test")
+	t.Logf("Resolving via upstream server on [TCP] --> %s", port)
+	addrs, err := dnsLookup(t, port, "tcp", "google.ca")
```
mook-as: Not really a fan of having a network dependency, but I guess standing up a stub DNS server isn't worth it.

pkg/commands/start_test.go
```diff
@@ -16,67 +16,84 @@ package commands
...
 	require.NoError(t, err, "Lookup IP failed")
+	require.True(t, len(addrs) > 0, true, "Expect at least an address")
```
mook-as: Should use `require.NotEmpty` here instead.

pkg/commands/start_test.go
```diff
@@ -16,67 +16,84 @@ package commands
...
 
-	expected := []net.IP{net.IPv4(222, 222, 222, 222)}
-	require.ElementsMatch(t, ipToString(addrs), ipToString(expected))
+	t.Logf("Resolving via upstream server on [UDP] --> %s", port)
+	addrs, err = dnsLookup(t, port, "udp", "google.ca")
+	require.NoError(t, err, "Lookup IP failed")
+	require.True(t, len(addrs) > 0, true, "Expect at least an address")
```
mook-as: `require.NotEmpty` here too.

mook-as (changes requested):

pkg/dns/dns.go
Nino-K: ha, I had a test case for this but I ended up removing it because the docker is always started with `--dns 127.0.0.1` this is due to the fact you cannot edit the `/etc/resolv.conf` once the container started. I ended up removing that test since there is no upstream DNS server so it is pointless to test.

pkg/commands/start_test.go
Nino-K: Yeah, not needed at all, it's an artifact of previous attempts.

pkg/commands/start.go
Nino-K: The funny thing is I actually prefer the way you suggested it, but everytime I do that other folks be like "can we put that into a variable" 😹

pkg/commands/start_test.go
```diff
@@ -106,6 +125,10 @@ func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]n
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
 
+func randomPort() string {
+	return fmt.Sprint(rand.Intn(32767-1024) + 1024)
```
mook-as: What you have is acceptable; but consider using:
```suggestion
	return fmt.Sprint(rand.Intn(32767) + 32768)
```
(i.e. some number between 32768 to 65535, inclusive)

mook-as (approved):

mook-as (approved):

pkg/commands/start_test.go
```diff
@@ -106,6 +125,10 @@ func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]n
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
 
+func randomPort() string {
+	return fmt.Sprint(rand.Intn(32767) + 32768)
+}
+
```
jandubois, line 131: This does not guarantee that the port is not currently in use. Please use the mechanism from https://github.com/lima-vm/lima/blob/3aca16d2a762fdb1ba9f50d6749cb03240d0a963/pkg/hostagent/hostagent.go#L181-L223 to pick a random available port.

jandubois (changes requested):
Random port should be a random available port

pkg/commands/start_test.go
Nino-K: @jandubois I have a bigger PR coming up where it implements the random ports if the user does not specify, can we merge this in the changes you suggested will be there.

jandubois (approved):

PR #7, opened 2022-04-12
Enables vsock between host and the peer

Nino-K:


.gitignore
```diff
@@ -0,0 +1,3 @@
+.DS_Store
+.vscode
```
jandubois: These patterns are not specific to this project and should not be in the repo `.gitignore`. You should put them into your `~/.gitignore` file instead.

cmd/host_windows.go
```diff
@@ -0,0 +1,58 @@
...
+var (
+	hostCmd = &cobra.Command{
+		Use:   "vsock-host",
+		Short: "Host-resolver vsock-host process",
+		Long: `Vsock-host runs the host-resolver as a host for AF_VSOCK in a host machine.
+It handles and forwards both TCP and UDP DNS queries over a virtual socket for the peer
+that runs inside a VM.
+
+--------------------HOST-------------------------------------WSL DISTRO------------
+| vsock-host | <----- AF_VOSOCK -----> [ VM ] <----- AF_VOSOCK -----> | vsock-peer |
```
mook-as:
```suggestion
| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
```

cmd/host_windows.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+			return commands.StartVsockHost(ipv6, hosts, upstreamServers)
+		},
+	}
+)
+
+func init() {
+	hostCmd.Flags().BoolP("ipv6", "6", false, "Enable IPv6 address family.")
+	hostCmd.Flags().StringToStringP("built-in-hosts", "c", map[string]string{},
+		"List of built-in Cnames to IPv4, IPv6 or IPv4-mapped IPv6 in host.rd.internal=111.111.111.111,com.backend.process=2001:db8::68 format.")
```
mook-as:
```suggestion
		"List of built-in CNAMEs to IPv4, IPv6 or IPv4-mapped IPv6 in host.rd.internal=111.111.111.111,com.backend.process=2001:db8::68 format.")
```

mook-as: `com.backend.process` is… misleading; that's a TLD of `.process`, right? It reads like it might be Java-style TLD-first instead…
Maybe use `example-host.rancherdesktop.io` instead. (Avoid the `example.` prefix, as `.example` is a TLD for documentation purposes.)

cmd/peer_linux.go
```diff
@@ -0,0 +1,71 @@
...
+// peerCmd represents the peer command
+var (
+	peerCmd = &cobra.Command{
+		Use:   "vsock-peer",
+		Short: "Host-resolver vsock-peer process",
+		Long: `Peer process to handle incoming TCP/UDP DNS queries over the
+AF_VSOCK connections from inside of the WSL VM. It is also a stub DNS forwarder for the host-resovler.
+
+--------------------HOST-------------------------------------WSL DISTRO------------
+| vsock-host | <----- AF_VOSOCK -----> [ VM ] <----- AF_VOSOCK -----> | vsock-peer |
```
mook-as:
```suggestion
| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
```

cmd/peer_linux.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
+)
+
+const defaultPort = 53
+
+// peerCmd represents the peer command
+var (
+	peerCmd = &cobra.Command{
+		Use:   "vsock-peer",
```
mook-as, line 32: Please add a top-level README or something similar to describe the peer process. While I understand what it is now, I'll probably forget in a month… :D

cmd/peer_linux.go
```diff
@@ -0,0 +1,71 @@
...
+		Short: "Host-resolver vsock-peer process",
+		Long: `Peer process to handle incoming TCP/UDP DNS queries over the
+AF_VSOCK connections from inside of the WSL VM. It is also a stub DNS forwarder for the host-resovler.
+
+--------------------HOST-------------------------------------WSL DISTRO------------
+| vsock-host | <----- AF_VOSOCK -----> [ VM ] <----- AF_VOSOCK -----> | vsock-peer |
+-----------------------------------------------------------------------------------`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// we will always listen for handshake connections from the host (server) incase of restarts
```
mook-as:
```suggestion
			// We will always listen for handshake connections from the host (server) in case of restarts.
```

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/Microsoft/go-winio"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows/registry"
+)
+
+const timeout = 10
+
+func vmGuid() (hvsock.GUID, error) {
```
mook-as, line 35: Please document what this function does.

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+	key, err := registry.OpenKey(
+		registry.LOCAL_MACHINE,
+		`SOFTWARE\Microsoft\Windows NT\CurrentVersion\HostComputeService\VolatileStore\ComputeSystem`,
+		registry.ENUMERATE_SUB_KEYS)
+	if err != nil {
+		return hvsock.GUIDZero, fmt.Errorf("could not retrieve registry key, is WSL VM running? %w", err)
```
mook-as:
```suggestion
		return hvsock.GUIDZero, fmt.Errorf("could not open registry key, is WSL VM running? %w", err)
```

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"time"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows/registry"
+)
+
+const timeout = 10
```
mook-as: Can we call this something like `timeoutSeconds`?

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+			if string(seed) == SeedPhrase {
+				log.Infof("successfully estabilished a handshake with a peer: %s", vmGuid.String())
+				found <- vmGuid
+			}
+			if err := conn.Close(); err != nil {
+				log.Errorf("hosthandshake closing connection: %v", err)
+			}
+		case <-bailOut:
+			log.Fatalf("all attempt to find a peer on WSL VM [%s] failed, is WSL or Peer running?", vmGuid.String())
```
mook-as: If there's 2 VMs (one WSL, one Hyper-V), the Hyper-V one will eventually hit bailOut and cause the whole process to die? That seems wrong.

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+	attempt := 1
+	for {
+		select {
+		case <-quit:
+			log.Debugf("attempt to handshake with [%s], goroutine is terminated", vmGuid.String())
+			return
+		case <-attempInterval.C:
+			conn, err := hvsock.Dial(addr)
+			if err != nil {
+				log.Errorf("attempt[%d] to handshake failed: %v", attempt, err)
```
mook-as: This may not be an error if this is the wrong VM… (That is, `log.Errorf` may be the wrong level to log at.)

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+				attempt++
+				continue
+			}
+			seed := make([]byte, len(SeedPhrase))
+			_, err = conn.Read(seed)
+			if err != nil {
+				log.Errorf("hosthandshake attempt to read the seed: %v", err)
+				return
+			}
+			if string(seed) == SeedPhrase {
```
mook-as: What if we got a partial read?

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,104 @@
...
+	defer close(done)
+
+	for _, name := range names {
+		vmGuid, err := hvsock.GUIDFromString(name)
+		if err != nil {
+			log.Errorf("invalid VM name: [%s], err: %v\n", name, err)
+			continue
+		}
+
+		go hostHandshake(vmGuid, found, done)
```
mook-as: If the WSL VM isn't up yet (at the point we fetch the names), we will just hang forever?

pkg/vmsock/defaults.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vmsock
+
+const (
+	HostListenPort    = uint32(5995)
+	PeerHandshakePort = uint32(7878)
```
mook-as: Given that this file is "defaults", shouldn't there be an override? (Or just rename the file "constants" instead…)

cmd/run.go
```diff
@@ -29,8 +29,12 @@ var (
 	runCmd = &cobra.Command{
 		Use:   "run",
 		Short: "Runs the host-resolver with a given arguments",
+		Long: `Runs the host-resolver in a server mode, this mode allows the host-resolver to
+attached to a defined IP and ports with given options. This mode is ideal for testing the contract
+with the underlying DNS server. Use this more for testing, debugging and benchmarks.
```
mook-as:
```suggestion
		Long: `Runs the host-resolver in a server mode, this mode allows the host-resolver to
be attached to a defined IP and ports with given options. This mode is ideal for testing the contract
with the underlying DNS server. Use this more for testing, debugging and benchmarks.
```

pkg/commands/standalone_server_test.go
```diff
@@ -129,9 +125,15 @@ func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]n
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
 
-//nolint:gosec
-func randomPort() string {
-	return fmt.Sprint(rand.Intn(32767) + 32768)
+// nolint: gocritic
```
mook-as: Please ensure the comment describes _which_ lint is being suppressed. That is, `gocritic` contains many lint rules; it's unclear which one it's violating.

If this is [`unamedresult`](https://go-critic.com/overview#unnamedresult), perhaps we should turn it off — it's off by default, and it doesn't sound very useful. I don't know why we turned it on in the first place.

pkg/commands/start_standalone.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	l, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		return 0, err
+	}
+	defer l.Close()
+	return l.Addr().(*net.TCPAddr).Port, nil
```
mook-as, line 63: Does that actually work? Wouldn't the port be in `CLOSED_WAIT` and the server would be unable to listen on it (until some time out)?

pkg/commands/start_host_windows.go
```diff
@@ -14,36 +14,29 @@ limitations under the License.
...
 
 import (
-	"os"
-	"os/signal"
-	"syscall"
-
 	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
-	"github.com/sirupsen/logrus"
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/vmsock"
+	log "github.com/sirupsen/logrus"
```
mook-as, line 19: Isn't this package normally imported as `logrus`? Why are we renaming it (making the use less clear)?

pkg/commands/start_standalone.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
+	log "github.com/sirupsen/logrus"
+)
+
+func StartStandAloneServer(address string,
+	udpLocalPort,
+	tcpLocalPort int,
+	ipv6 bool,
+	hosts map[string]string,
+	upstreamServers []string) error {
```
mook-as: That's a lot of arguments — consider using an options object instead. In fact, all the arguments are in `dns.ServerOptions`…

pkg/dns/dns.go
```diff
@@ -282,6 +282,21 @@ func (h *Handler) ServeDNS(w dns.ResponseWriter, req *dns.Msg) {
...
 	}
 }
+func StartWithListener(opts ServerOptions) (*Server, error) {
+	h, err := newHandler(opts.IPv6, opts.StaticHosts, opts.UpstreamServers)
+	if err != nil {
+		return nil, err
+	}
+	server := &Server{}
+	s := &dns.Server{Net: "tcp", Listener: opts.Listener, Handler: h}
+	server.tcp = s
```
mook-as: Why do we need the extra assignment?
```suggestion
	s := &dns.Server{Net: "tcp", Listener: opts.Listener, Handler: h}
	server := &Server{tcp: s}
```

pkg/dns/dns.go
```diff
@@ -282,6 +282,21 @@ func (h *Handler) ServeDNS(w dns.ResponseWriter, req *dns.Msg) {
 		h.handleDefault(w, req)
 	}
 }
+func StartWithListener(opts ServerOptions) (*Server, error) {
```
mook-as: It would be good if we add some documentation to this function:
- Note that the "Listener" in the function name refers to `opts.Listener`
- Say something about this always being treated as TCP (as opposed to UDP) for the DNS server.

mook-as: Nit: should have empty line before the function (at least the other functions in the file have it).

pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,112 @@
...
+		log.Errorf("handleUDP write to original requester: %v", err)
+	}
+}
+
+func appendHeaderLen(m []byte, n int) []byte {
+	// This is to accommodate for the header length that exists in TCP buffer for DNS payload
+	// DNS server does a binary read on the first 8 byte(uint16) of the buffer for this length
+	// the UDP buffer does not include this header length, therefore it would need to be added
+	msg := make([]byte, 2+n)
+	// we cannot do a bindary.Write on net.Con since this is a UDP Conn
```
mook-as:
```suggestion
	// we cannot do a binary.Write on net.Conn since this is a UDP Conn
```

pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,112 @@
...
+	_, err = uConn.WriteToUDP(data[2:], addr)
+	if err != nil {
+		log.Errorf("handleUDP write to original requester: %v", err)
+	}
+}
+
+func appendHeaderLen(m []byte, n int) []byte {
+	// This is to accommodate for the header length that exists in TCP buffer for DNS payload
+	// DNS server does a binary read on the first 8 byte(uint16) of the buffer for this length
+	// the UDP buffer does not include this header length, therefore it would need to be added
```
mook-as: This should be a doc comment (_above_ the function declaration).

pkg/vmsock/conn_linux.go
mook-as: This function doesn't have a `Conn` at all to write to?

pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Listen, could not parse Hyper-v service GUID: %v", err)
+	}
+
+	addr := hvsock.Addr{
+		VMID:      vmGuid,
+		ServiceID: svcPort,
+	}
+
+	return hvsock.Listen(addr)
```
mook-as, line 40: Does that actually work? I was under the impressions that, in order for Windows processes to listen on the HV_VSOCK, it needs to be [registered](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application).

mook-as (changes requested):

pkg/vmsock/host_handshake_windows.go
Nino-K: The seed is less than 50 chars and it is always a static string, TCP will not attempt to send over multiple segmentation. Using a buffer here will be an overkill.

pkg/commands/start_standalone.go
Nino-K: it works, I'm using it in all the tests.

pkg/commands/standalone_server_test.go
Nino-K: The `gocritic` is a useful one. I don't think we want that off. I left a comment for the reason which should be clear enough.

pkg/commands/start_host_windows.go
Nino-K: I honestly think log is more clear than logrus, but I'm ok either way, let me know if this is too crazy I will remove the rename 🤷

pkg/vmsock/conn_windows.go
Nino-K: Yes, I came across the docs, however, the registration step seems unnecessary.

cmd/host_windows.go
```diff
@@ -0,0 +1,61 @@
...
+	hostCmd = &cobra.Command{
+		Use:   "vsock-host",
+		Short: "Host-resolver vsock-host process",
+		Long: `Vsock-host runs the host-resolver as a host for AF_VSOCK in a host machine.
+It handles and forwards both TCP and UDP DNS queries over a virtual socket for the peer
+that runs inside a VM.
+
+--------------------HOST-------------------------------------WSL DISTRO------------
+| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
+-----------------------------------------------------------------------------------`,
```
mook-as: Oops, didn't notice that your chart isn't lined up anymore…
```suggestion
--------------------HOST------------------------------------WSL DISTRO------------
| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
----------------------------------------------------------------------------------`,
```

cmd/peer_linux.go
```diff
@@ -0,0 +1,74 @@
...
+var (
+	peerCmd = &cobra.Command{
+		Use:   "vsock-peer",
+		Short: "Host-resolver vsock-peer process",
+		Long: `Peer process to handle incoming TCP/UDP DNS queries over the
+AF_VSOCK connections from inside of the WSL VM. It is also a stub DNS forwarder for the host-resovler.
+
+--------------------HOST-------------------------------------WSL DISTRO------------
+| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
+-----------------------------------------------------------------------------------`,
```
mook-as:
```suggestion
--------------------HOST------------------------------------WSL DISTRO------------
| vsock-host | <----- AF_VSOCK -----> [ VM ] <----- AF_VSOCK -----> | vsock-peer |
----------------------------------------------------------------------------------`,
```

cmd/run.go
```diff
@@ -28,9 +31,20 @@ var (
 
 	runCmd = &cobra.Command{
 		Use:   "run",
-		Short: "Runs the host-resolver with a given arguments",
+		Short: "Runs the host-resolver standalone server with a given arguments",
+		Long: `Runs the host-resolver in a server mode, this mode allows the host-resolver to
+attach to a defined IP and ports with given options. This mode is ideal for testing the contract
+with the underlying DNS server. Use this mode for testing, debugging and benchmarks.
```
mook-as:
```suggestion
		Long: `Runs the host-resolver in standalone mode; this mode allows the host-resolver to
attach to a defined IP and ports with given options. This mode is ideal for testing the contract
with the underlying DNS server. Use this mode for testing, debugging and benchmarks.
```

cmd/run.go
```diff
@@ -28,9 +31,20 @@ var (
 
 	runCmd = &cobra.Command{
 		Use:   "run",
```
mook-as: Maybe renamed this to `run-standalone` or just `standalone` or something.

pkg/commands/standalone_server_test.go
mook-as: Your comment should include the word `unamedresult`, since that's the rule being ignored…

pkg/commands/start_host_windows.go
mook-as: I've mostly seen it used as `logrus` (because it has a slightly different API than the built-in [`log`](https://pkg.go.dev/log)), but I don't care enough to make you change it all :)

pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,41 @@
...
+	"net"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
+)
+
+func Listen() (net.Listener, error) {
+	vmGuid, err := vmGuid()
+	if err != nil {
+		return nil, fmt.Errorf("Listen, could not use VM GUID: %v", err)
```
mook-as:
```suggestion
		return nil, fmt.Errorf("Listen, could not determine VM GUID: %v", err)
```

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,120 @@
...
+			return vmGuid, nil
+		case <-bailOut:
+			return hvsock.GUIDZero, errors.New("could not find vsock-peer process on any hyper-v VM(s)")
+		}
+	}
+}
+
+// handshake attempts to perform a handshake by verifying the seed with a running
+// af_vsock peer in WSL distro, it attempts once per second
+func handshake(vmGuid hvsock.GUID, found chan hvsock.GUID, done chan bool) {
```
mook-as:
```suggestion
func handshake(vmGuid hvsock.GUID, found chan<- hvsock.GUID, done <-chan bool) {
```
No need to read from `found`, or write to `done`.

pkg/vmsock/host_handshake_windows.go
```diff
@@ -0,0 +1,120 @@
...
+	attempt := 1
+	for {
+		select {
+		case <-done:
+			log.Infof("attempt to handshake with [%s], goroutine is terminated", vmGuid.String())
+			return
+		case <-attempInterval.C:
+			conn, err := hvsock.Dial(addr)
+			if err != nil {
+				attempt++
```
mook-as, line 104: Should we `log.Debugf` or something so we can tell later that this happened (for debugging)?

pkg/vmsock/host_handshake_windows.go
mook-as: This is vsock, not TCP; but I can see that we might expect to get a full read. (I'd just use [`io.ReadAtLeast`](https://pkg.go.dev/io#ReadAtLeast) for this…)

mook-as (changes requested):

Nino-K:
> This is vsock, not TCP; but I can see that we might expect to get a full read. (I'd just use [io.ReadAtLeast](https://pkg.go.dev/io#ReadAtLeast) for this…)

Of course, it's not TCP but remember, it's still based on `stream_socket` API and we use all the same underlying methods :)

It's it updated now.

cmd/run.go
```diff
@@ -15,10 +15,13 @@ package cmd
 
 import (
 	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/commands"
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
 	"github.com/spf13/cobra"
 )
 
-// runCmd represents the run command
+// runCmd represent the run command, it startup a standalone server that listens on a given
```
mook-as: This is no longer the `run` command.

mook-as (changes requested):

mook-as (approved):

PR #9, opened 2022-04-12
Create go.yml

Nino-K:


.golangci.yml
```diff
@@ -0,0 +1,106 @@
+run:
+  skip-files:
+    - pkg/dns/dns.go
```
mook-as: Why are we skipping this file? We should ideally drop the exception, but at least add a comment.

.golangci.yml
```diff
@@ -0,0 +1,106 @@
...
+      - '1'
+      - '2'
+      - '3'
+    ignored-functions:
+      - strings.SplitN
+
+  lll:
+    line-length: 140
+  misspell:
+    locale: US
```
mook-as, line 46: Aww, not en_CA? :p

.golangci.yml
```diff
@@ -0,0 +1,106 @@
...
+  nolintlint:
+    allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
+    allow-unused: false # report any unused nolint directives
+    require-explanation: false # don't require an explanation for nolint directives
```
mook-as, line 50: Why not?

.golangci.yml
```diff
@@ -0,0 +1,106 @@
...
+  # - godot
+  # - godox
+  # - goerr113
+  # - interfacer
+  # - maligned
+  # - nestif
+  # - prealloc
+  # - testpackage
+  # - revive
+  # - wsl
```
mook-as: Nit: missing EOL at EOF.

.golangci.yml
```diff
@@ -0,0 +1,106 @@
...
+      - '2'
+      - '3'
+    ignored-functions:
+      - strings.SplitN
+
+  lll:
+    line-length: 140
+  misspell:
+    locale: US
+  nolintlint:
```
mook-as, line 47: Wait, this linter isn't even enabled, but we have configs for it?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.18.0-stretch
+FROM golang:latest
```
mook-as, line 1: Hmm, I tend to like pinning versions, but I can see the use in tracking latest too…

cmd/run.go
```diff
@@ -35,6 +35,7 @@ var (
 	}
 )
 
+//nolint
```
mook-as: Why no lint?

pkg/commands/start.go
```diff
@@ -22,16 +22,16 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func Start(address string, udpLocalPort, tcpLocalPort int, IPv6 bool, hosts map[string]string, upstreamServers []string) error {
+func Start(address string, udpLocalPort, tcpLocalPort int, iPv6 bool, hosts map[string]string, upstreamServers []string) error {
```
mook-as: That seems wrong — should probably be `ipv6` or something?

mook-as (changes requested):
Should turn on the lint to say we need to have a reason for each nolint.
(Unless you want to do follow-up PRs; in that case, maybe make a PR to fix things, before a second PR to turn on linting.)

.golangci.yml
Nino-K: This file belongs to upstream, and eventually will not be part of this project.

.golangci.yml
Nino-K: I don't believe that's a thing. The available options are UK or US. Canada is closer to US spelling than UK.

.golangci.yml
Nino-K: I personally think it would make a mess in the codebase. One option is to move the explanations to the golanci.yml file.

.golangci.yml
Nino-K: the linter is enabled, and the `nolintlint` is a linter check on `nolint` inline comments. It conflicts with gopls gofmt currently. 

I see, I left the rules behind when I disabled it.

cmd/run.go
Nino-K: Linter has a rule for init functions `gochecknoinits` which is annoying. Probably should figure out something better since all the cobra cli stuff is built around `init funcs`. Maybe I can update to `//nolint:gochecknoinits` so we just skip that one rule.

pkg/commands/start.go
Nino-K: yeah, that looks wrong

.golangci.yml
Nino-K: Yeah, I don't think the support anything other than UK and US https://github.com/istio/istio/blob/master/.gitignore

.golangci.yml
Nino-K: Yeah, I don't think they support anything other than UK and US https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/misspell.go#L86

Nino-K:
> Should turn on the lint to say we need to have a reason for each nolint.

@mook-as I think the explanation is nice to have(voluntarily as opposed to a rule) where things are confusing and linter has no specific rules, I believe having an explanation on all the `nolints` could be really pointless e.g. consider the scenario below: 

```
package cobraCliStuff

//nolint:gochecknoinits 
func init() {

}
```

What is a meaningful explanation? the nolint with a corresponding rule can be pretty helpful IMO. Otherwise, the explanation has to be something really vague like:

`//nolint:gochecknoinits \\ cobra cli loves init funcs` 🤷 

I'm ok, to turn on the explanations, I'm just going based on my experiences in the past where we had the explanation on and never found it helpful, just a clutter. 

On the other hand, I think all the //nolints should have follow-up rules.

.golangci.yml
mook-as: I agree that `gochecknoinits` doesn't need explanations (in which case, [`allow-no-explanation`](https://golangci-lint.run/usage/linters#nolintlint) would be useful); I'm not sure yet whether we'll end up with too many lints with obvious reasons… maybe we can come back to this later.

.golangci.yml
mook-as: (This is still relevant — `.golangci.yml` is still missing EOL at EOF.)

.golangci.yml
```diff
@@ -0,0 +1,105 @@
...
+  enable:
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - dupl
+    - errcheck
+    - exportloopref
+    - funlen
+    - gochecknoinits
```
mook-as, line 64: Given that we need `init()`s for cobra, does it make sense to turn this on?

Maybe we can exclude this one from `cmd/*.go`? (Via [`exclude-rules`](https://golangci-lint.run/usage/configuration/#issues-configuration).)

mook-as (changes requested):

.golangci.yml
Nino-K: Sure, I think we can fine-tune this further later on.

.golangci.yml
Nino-K: sure, I think it makes sense to disable `gochecknoinits` that in the `cmd` stuff.

.golangci.yml
```diff
@@ -0,0 +1,112 @@
...
+  # - testpackage
+  # - revive
+  # - wsl
+
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    - path: cmd/*
+      linters:
+        - gochecknoinits
```
mook-as, line 112: Would be to have a comment (because it's not obvious from the config file that this is for cobra).

pkg/commands/start_test.go
```diff
@@ -125,6 +129,7 @@ func dnsLookup(t *testing.T, resolverPort, resolverProtocol, domain string) ([]n
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
 
+//nolint:gosec
```
mook-as, line 132: Do you recall _why_ `gosec` thought this was an issue?

mook-as (changes requested):

pkg/commands/start_test.go
Nino-K: This is removed in the bigger PR and should no longer be the case. Basically, linter doesn't like weak random number generators which doesn't make sense in this case.

mook-as: Sounds good.

mook-as (approved):

PR #11, opened 2022-04-21
Add release matrix

Nino-K:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  releases-matrix:
+    name: Release Binaries
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # build and publish in parallel: linux/386, linux/amd64, linux/arm64, windows/386, windows/amd64, darwin/amd64, darwin/arm64
+        goos: [linux, windows, darwin]
+        goarch: ["386", amd64, arm64]
```
mook-as: I don't think we actually support `386` anywhere.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,29 @@
...
+        exclude:
+          - goarch: "386"
+            goos: darwin
+          - goarch: arm64
+            goos: windows
+    steps:
+    - uses: actions/checkout@v2
+    - uses: wangyoucao577/go-release-action@v1.26
```
mook-as: Please pin to hash for non-official actions (`actions/`, docker ones for docker, etc.)

Of course, ideally we'd _not_ use third party actions, and just do `setup-go` + build + `gh` or alternative.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        goos: ${{ matrix.goos }}
+        goarch: ${{ matrix.goarch }}
+        goversion: 1.18
+        binary_name: "resolver"
+        extra_files: LICENSE README.md
```
mook-as, line 27: Do I understand correctly that we're uploading these files multiple times? (That's okay, just surprising)

mook-as (changes requested):

mook-as (approved):

PR #13, opened 2022-04-26
build with cgo disabled

Nino-K:


Nino-K:
I guess the env can also be declared hight up perhaps after `on:`

mook-as (approved):
Shouldn't hurt — as far as we can tell DNS on Windows is enabled regardless of CGO?

PR #14, opened 2022-05-03
match the length of data to the slice when returning

Nino-K:
Fixes the following issue: https://github.com/rancher-sandbox/rancher-desktop/issues/2141

Busybox nslookup expects the length of received data to match the array the length of the array that is received in, meaning that no padding is accepted. 

https://elixir.bootlin.com/busybox/latest/source/networking/nslookup.c#L640

### DIFF
--- a/pkg/vmsock/conn_linux.go
+++ b/pkg/vmsock/conn_linux.go
@@ -24,7 +24,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const UDPMaxBufer = 1024
+const UDPMaxBuffer = 512
 
 func ListenTCP(addr string, port int) error {
 	l, err := net.ListenTCP("tcp", &net.TCPAddr{IP: net.ParseIP(addr), Port: port})
@@ -64,7 +64,7 @@ func ListenUDP(addr string, port int) error {
 	}
 	defer l.Close()
 	for {
-		buf := make([]byte, UDPMaxBufer)
+		buf := make([]byte, UDPMaxBuffer)
 		n, addr, err := l.ReadFromUDP(buf)
 		if err != nil {
 			log.Errorf("ListenUDP, read error: %v", err)
@@ -87,14 +87,21 @@ func handleUDP(uConn *net.UDPConn, addr *net.UDPAddr, b []byte) {
 		log.Errorf("handleUDP write to vsock host: %v", err)
 		return
 	}
-	data := make([]byte, UDPMaxBufer)
-	_, err = conn.Read(data)
+	data := make([]byte, UDPMaxBuffer)
+	n, err := conn.Read(data)
 	if err != nil {
 		log.Errorf("handleUDP read from vsock host: %v", err)
 		return
 	}
-	// remove the tcp length from the beginning since that is the place for msg id ;b
-	_, err = uConn.WriteToUDP(data[2:], addr)
+	// TCP DNS requests have two-byte length prefix, which is not present in UDP
+	// DNS requests (See RFC 1035).  Drop the prefix; however, note that the
+	// length of the write needs to match the length of the received slice (as
+	// opposed to its capacity).  This is necessary since some tools such as
+	// busybox nslookup explicitly check this.
+	// https://git.busybox.net/busybox/tree/networking/nslookup.c?h=1_35_0#n640
+	reqBody := data[2:n]
+	_, err = uConn.WriteToUDP(reqBody, addr)
+
 	if err != nil {
 		log.Errorf("handleUDP write to original requester: %v", err)
 	}


pkg/vmsock/conn_linux.go
```diff
@@ -88,13 +88,19 @@ func handleUDP(uConn *net.UDPConn, addr *net.UDPAddr, b []byte) {
...
 		return
 	}
-	// remove the tcp length from the beginning since that is the place for msg id ;b
-	_, err = uConn.WriteToUDP(data[2:], addr)
+	// data[2:n] is to remove the tcp length from the beginning
+	// since that is the place for msg id
+	// IMPORTANT: the write needs to match the length of the data to
+	// the length of the returned slice. This is necessary since some
+	// tools such as busybox nslookup they explicitly check this
+	// https://elixir.bootlin.com/busybox/latest/source/networking/nslookup.c#L640
```
mook-as: We should always pin to a commit (and never `latest`); also, that isn't the canonical repository.
Should probably be https://git.busybox.net/busybox/tree/networking/nslookup.c?h=1_35_0#n640 instead?
It's also unclear to me how that line of source maps to the comment, but okay…

pkg/vmsock/conn_linux.go
```diff
@@ -88,13 +88,19 @@ func handleUDP(uConn *net.UDPConn, addr *net.UDPAddr, b []byte) {
...
 	data := make([]byte, UDPMaxBufer)
-	_, err = conn.Read(data)
+	n, err := conn.Read(data)
 	if err != nil {
 		log.Errorf("handleUDP read from vsock host: %v", err)
 		return
 	}
-	// remove the tcp length from the beginning since that is the place for msg id ;b
-	_, err = uConn.WriteToUDP(data[2:], addr)
+	// data[2:n] is to remove the tcp length from the beginning
```
mook-as: I find the wording a bit unclear; maybe have a temporary variable to attach the comment to?
```go
	// TCP DNS requests have two-byte length prefix, which is not present in UDP
	// DNS requests (See RFC 1035).  Drop the prefix; however, note that the
	// length of the write needs to match the length of the received slice (as
	// opposed to its capacity).  This is necessary since some tools such as
	// busybox nslookup explicitly check this.
	reqBody := data[2:n]
	_, err = uConn.WriteToUDP(reqBody, addr)
```

pkg/vmsock/conn_linux.go
```diff
@@ -88,13 +88,19 @@ func handleUDP(uConn *net.UDPConn, addr *net.UDPAddr, b []byte) {
 		return
 	}
 	data := make([]byte, UDPMaxBufer)
```
mook-as:
```suggestion
	data := make([]byte, UDPMaxBuffer)
```
We totally missed the typo…

pkg/vmsock/conn_linux.go
```diff
@@ -88,13 +88,19 @@ func handleUDP(uConn *net.UDPConn, addr *net.UDPAddr, b []byte) {
...
-	_, err = conn.Read(data)
+	n, err := conn.Read(data)
```
mook-as, line 91: Given that the data has a two byte length prefix, shouldn't we check that `n` is at least `data[0:1]` bytes long?

pkg/vmsock/conn_linux.go
mook-as: Actually, given that the first two bytes of `data` is a length, we should use that instead of the return value of `Read`.  This can matter if we have an over-read.

pkg/vmsock/conn_linux.go
mook-as: Hmm, but this will also mean we need to sanity-check it (though that's 64k max, so it should be fine… other than the fact that the other end is UDP, so we need to cap it at 512).

mook-as (changes requested):
Note that this seems to break for long responses; check the `TXT` record for `suse.com` for an example. (I get 959 bytes with 12 answers without host-resolver, but an empty response with.)

pkg/vmsock/conn_linux.go
jandubois: I think it is actually the code in [ns_parse.c](https://github.com/bminor/musl/blob/7a43f6fea9081bdd53d8a11cef9e9fab0348c53d/src/network/ns_parse.c#L104) from the musl library that is more relevant, but it doesn't really matter...

Random bonus tip (not applicable here): For repos on Github you can press the `y` key after selecting some lines in a file, and the URL will change to the latest commit that has modified those lines.

pkg/vmsock/conn_linux.go
jandubois: > the other end is UDP, so we need to cap it at 512

Isn't that an application layer concern. The tunnel should just pass through the data as it receives it (and it should already be truncated by the DNS server). The `vmsock` code shouldn't know anything about DNS internals.

Nino-K:
> Note that this seems to break for long responses; check the `TXT` record for `suse.com` for an example. (I get 959 bytes with 12 answers without host-resolver, but an empty response with.)

This change should have not impacted the long `TXT` resource record responses. This change only applies to UDP and anything longer than `512` is either upgraded to `eDNS` or `TCP`. I will create a separate issue to see why host-resolver is not upgrading to `TCP` for long responses.

pkg/vmsock/conn_linux.go
Nino-K: Sure, I always use `y` if the link is on github to get the canonical form, however, I could not find the source link on github.

pkg/vmsock/conn_linux.go
mook-as: No, because our upstream (vsock) is a stream-based DNS server, and doesn't have the 512 byte limit.  And upstream of that is some syscalls (well, Win32 API calls), and has no limit either.

pkg/vmsock/conn_linux.go
Nino-K: > This can matter if we have an over-read.

@mook-as could you elaborate? I'm trying to understand how we can have an over-read in UDP segments?

Nino-K:
Quick update, the txt records longer than 512 or anything longer than 512 bytes (on TCP) will not work currently since we are truncating everything down to 512 bytes [here](https://github.com/rancher-sandbox/rancher-desktop-host-resolver/blob/e5c88519ed51e4ce64502ebc34ad43f1de5f8dd4/pkg/dns/dns.go#L252), which is completely wrong. 

I will submit a separate PR to address this issue.

pkg/vmsock/conn_linux.go
mook-as: You're reading vsock, not UDP (but _writing_ to UDP), right?

pkg/vmsock/conn_linux.go
Nino-K: So here is how I look at it if there is no error that means we have received a UDP response. Therefore, there is no need to check the `n` because we know we have received a UDP datagram, and the first 16 bit or 2 octets (fixed size value) are guaranteed to be the UDP datagram length in the header (that's of course in a Big Endian system which is the entire internet). Also, we know there will be no padding involved in the header.

pkg/vmsock/conn_linux.go
Nino-K: We are reading `[]byte` that coming out of vsock and the slice of byte represent a UDP datagram, I believe the vsock is just a form of transport or connection to the other end that talks UDP. The connection itself is the same object that is shared with `net.Conn` or `net.UDPConn` which is independent of the bytes themselves.  Hence, the original caller such as nslookup, dig, or any other DNS client knows nothing about vsock.

mook-as: Discussed with Nino: the truncation stuff is going to have to be done via two AF_VSOCK connections to the host, as we should only truncate on UDP, but we don't know if a request is UDP (and there's no good way for us to figure that out by the time we should truncate).

This will need to be done in a separate PR.

mook-as (approved):

pkg/vmsock/conn_linux.go
jandubois: This is wrong; the transport layer should not have any knowledge about the application layer protocols.

Now it doesn't matter right now, since we make sure the packets are small enough, but conceptually it still looks wrong to me.

PR #15, opened 2022-05-04
don't look for /etc/resolv.conf on windows

Nino-K:
This could be misleading when the host process looks for `/etc/resolv.conf` on the windows and emits a log line that can be misleading. Also, we should set the fallback IPs if the upstream assignment fails. 

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -5,6 +5,7 @@ package dns
 import (
 	"fmt"
 	"net"
+	"runtime"
 	"strings"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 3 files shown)


pkg/dns/dns.go
```diff
@@ -70,9 +71,16 @@ func newHandler(IPv6 bool, hosts map[string]string, upstreamServers []string) (d
...
+			cc, err = dns.ClientConfigFromFile("/etc/resolv.conf")
+			if err != nil {
+				logrus.WithError(err).Warnf("failed to detect system DNS, falling back to %v", defaultFallbackIPs)
+				cc, err = newStaticClientConfig(defaultFallbackIPs)
+				if err != nil {
+					return nil, err
+				}
+			}
+		} else {
 			cc, err = newStaticClientConfig(defaultFallbackIPs)
```
mook-as, line 86: Should probably add a comment saying that on Windows, we only ever use hard-coded servers for fallback.

pkg/dns/dns.go
```diff
@@ -81,7 +89,10 @@ func newHandler(IPv6 bool, hosts map[string]string, upstreamServers []string) (d
 	} else {
 		cc, err = newStaticClientConfig(upstreamServers)
 		if err != nil {
-			return nil, err
+			cc, err = newStaticClientConfig(defaultFallbackIPs)
+			if err != nil {
+				return nil, err
+			}
 		}
```
mook-as: (Optional) Maybe do it like
```go
if cc, err = … ; err != nil {
  if cc, err = … ; err != nil {
    return nil, err
```

mook-as (changes requested):
Checked with `godbolt.org` — the default compiler does optimize out code based on `runtime.GOOS`.

pkg/dns/dns.go
```diff
@@ -70,18 +71,28 @@ func newHandler(IPv6 bool, hosts map[string]string, upstreamServers []string) (d
...
 			if err != nil {
 				return nil, err
 			}
 		}
 	} else {
-		cc, err = newStaticClientConfig(upstreamServers)
-		if err != nil {
-			return nil, err
+		if cc, err = newStaticClientConfig(upstreamServers); err != nil{
+			if cc, err = newStaticClientConfig(defaultFallbackIPs); err != nil{
```
mook-as: Nit: space before brace.

mook-as (approved):

pkg/dns/dns.go
Nino-K: How did that pass the linter 🤔

mook-as: https://github.com/rancher-sandbox/rancher-desktop-host-resolver/blob/325fb3dc92af0804f94c8e2c0da3d5e6f195ba21/.golangci.yml#L1-L3

mook-as (approved):

PR #16, opened 2022-05-04
add viper env make e2e easier

Nino-K:
This will allow both host and peer processes to receive env var during start-up which can make e2e testing easier. 


mook-as (changes requested):
(Needs to be in standalone too so we test _something_.)

cmd/run.go
```diff
@@ -37,28 +36,39 @@ attach to a defined IP and ports with given options. This mode is ideal for test
 with the underlying DNS server. Use this mode for testing, debugging and benchmarks.
 		`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			allSettings := runViper.AllSettings()
+			builtInHosts := allSettings["built-in-hosts"]
+			hosts, err := cast.ToStringMapStringE(builtInHosts)
+			if err != nil {
+				logrus.Errorf("reading built-in-hosts value: %v", err)
+			}
```
mook-as:
```suggestion
			hosts := runViper.GetStringMapString("built-in-hosts")
```
Maybe I'm missing something?

cmd/host_windows.go
```diff
@@ -35,18 +40,16 @@ that runs inside a VM.
...
+			ipv6 := hostViper.GetBool("ipv6")
+			upstreamServers := hostViper.GetStringSlice("upstream-servers")
+
+			allSettings := hostViper.AllSettings()
+			builtInHosts := allSettings["built-in-hosts"]
+			hosts, err := cast.ToStringMapStringE(builtInHosts)
 			if err != nil {
-				return err
+				logrus.Errorf("reading built-in-hosts value: %v", err)
 			}
```
mook-as:
```suggestion
			hosts := runViper.GetStringMapString("built-in-hosts")
```
(Again, I might be missing something…)

mook-as (changes requested):
Just some API use questions; the bulk of the PR looks fine.

cmd/run.go
Nino-K: I tried, but it doesn't work. I think viper has issues with receiving maps through env-var.

mook-as: Odd; it seemed to work for me (or at least, the tests still pass). That was using `StaticHosts:     runViper.GetStringMapString("built-in-hosts"),`.
(I initially forgot to remove the `cast` import and the build failed, so I'm definitely making changes to the right file at least…)

Of course, there's also the chance that the test isn't working correctly, which is an issue too.

Nino-K: I see what happened, it was actually the serialization while setting the value in the test that fixed the issue not casting. When I set the data I didn't get to test it without the casting stuff. It is working as expected. I also, confirmed that the test fails if values are not set to expected values.

mook-as (approved):

PR #18, opened 2022-05-09
Add-test-dns-server

Nino-K:
- This will need additional Resource Records other than A records

This test server will run alongside the host-resolver host process on the windows machine. It will act as the upstream DNS server for the host-resolver Host process. This is to eliminate the dependency on the internet during the e2e test.

Please note, currently, A records are being handled only, however, this will change to include additional resource records.

The test DNS server can run as follows:
```golang
testServer := Server{
         Addr: "127.0.0.1", 
         TCPPort: "53",
         UDPPort: "53",
}

testServer.Run()
```


pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+						},
+						A: net.ParseIP(addr),
+					})
+				}
+			}
+		}
+	}
+}
+
+func loadRecords(path string) map[string][]string {
```
mook-as: Is there any point in using actualy DNS records? Or should we just make up things (like `1.host.example`)?

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+		var ips = []string{}
+		for j, field := range line {
+			switch j {
+			case 0:
+				domain = fmt.Sprintf("%s.", field)
+			case 1:
+				ips = append(ips, field)
+			case 2:
+				ips = append(ips, field)
+			}
```
mook-as:
```suggestion
			if j == 0 {
				domain = fmt.Sprintf("%s.", field)
			} else {
				ips = append(ips, field)
			}
```
This doesn't make sense as a switch (and using it also limits it to two addresses for no reason).

pkg/test/data/test-300.csv
```diff
@@ -0,0 +1,300 @@
...
+www.microsoft.com,24.110.62.247,43.70.157.107
+o.lencr.org,63.110.16.26,127.234.14.182
+sfx.ms,164.123.67.6,11.165.236.190
+asm.skype.com,91.55.129.128,123.211.200.31
+amer.ng.msg.teams.microsoft.com,21.250.161.213,165.253.193.171
+oneclient.sfx.ms,252.18.29.101,67.11.113.58
+gsp64-ssl.ls.apple.com,44.47.211.139,24.12.153.253
+ups.analytics.yahoo.com,57.203.35.211,245.133.75.93
+ssl.google-analytics.com,155.100.41.29,80.138.79.122
+a.akamaiedge.net,87.8.39.235,7.133.87.69
```
mook-as: (Missing EOL at EOF)

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+				}
+			}
+		}
+	}
+}
+
+func loadRecords(path string) map[string][]string {
+	f, err := os.Open(path)
+	if err != nil {
+		log.Panicf("opening file: %v err: %v", path, err)
```
mook-as: The mix of `log` & `logrus` is a bit confusing. `logrus` has a `Panicf` too; should we just use that instead?

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	defer func() {
+		err := f.Close()
+		if err != nil {
```
mook-as:
```suggestion
		if err := f.Close(); err != nil {
```

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+
+const truncateSize = 512
+
+type Server struct {
+	Addr    string
+	TCPPort int
+	UDPPort int
+}
+
+func (s *Server) parseUDPAddr() string {
```
mook-as: There's no parsing involved here; maybe just call this `udpAddr()`? (Same with the TCP version.)
Or just call `net.JostHostPort()` directly and skip the wrapper method?

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+func (s *Server) parseTCPAddr() string {
+	return fmt.Sprintf("%v:%v", s.Addr, s.TCPPort)
+}
+
+func (s *Server) Run() {
+	g := new(errgroup.Group)
+	udpSrv := &dns.Server{Addr: s.parseUDPAddr(), Net: "udp", Handler: &udpHandler{}}
+	defer func() {
+		err := udpSrv.Shutdown()
+		if err != nil {
```
mook-as:
```suggestion
		if err := udpSrv.Shutdown(); err != nil {
```

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+			logrus.Errorf("shutting down UDP server: %v", err)
+		}
+	}()
+	g.Go(func() error {
+		return udpSrv.ListenAndServe()
+	})
```
mook-as: Can we get away with just calling the method directly?
```suggestion
	g.Go(udpSrv.ListenAndServe)
```

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			logrus.Errorf("shutting down TCP server: %v", err)
+		}
+	}()
+	g.Go(func() error {
+		return tcpSrv.ListenAndServe()
+	})
+
+	if err := g.Wait(); err != nil {
+		logrus.Fatalf("Failed starting servers: %v", err)
```
mook-as: This also happens if the servers started, and then later had an error, right?

pkg/test/server.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+type tcpHandler struct{}
```
mook-as: Consider merging the two handlers:
```go
type handler struct{
 	truncate bool
}
…
func (h *Handler) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
	…
	parseReply(msg)
	if h.truncate {
		msg.Truncate(truncateSize)
	}
	…
udpSrv := &dns.Server{Addr: s.parseUDPAddr(), Net: "udp", Handler: &handler{truncate: true}}
tcpSrv := &dns.Server{Addr: s.parseTCPAddr(), Net: "tcp", Handler: &handler{}}
```
```

File: pkg/test/server.go [hunk 11]
```diff
@@ -0,0 +1,156 @@
...
+	err := w.WriteMsg(msg)
+	if err != nil {
+		logrus.Errorf("ServeDNS trying to write message in udpHandler: %v", err)
+	}
+}
+
+func parseReply(msg *dns.Msg) {
+	logrus.Infof("ServeDNS handling request: %v", msg.Question)
+	for _, q := range msg.Question {
+		switch q.Qtype { //nolint:gocritic
```
mook-as: Please describe why we're suppressing the lint. (Is it because we're intentionally not covering all cases? Because we have a single-clause `switch`?)

mook-as (changes requested):
Please have something that describes how this will be used; currently it's just code sitting there with no callers. (In the PR description is fine, if we expect to use it shortly.)

pkg/test/server.go
Nino-K: Not too sure at this point which path we will take, either static data or generating data on the fly. For now, we only have A records but that will change shortly as we need to add other RRs. Maybe, we can make this decision then since it is a simple change.

pkg/test/server.go
Nino-K: True, I had another column initially. I will cleanup.

pkg/test/server.go
```diff
@@ -0,0 +1,125 @@
...
+	// we might handle other OpCodes later
+	if r.Opcode != dns.OpcodeQuery {
+		return
+	}
+	parseReply(msg)
+	if h.truncate {
+		msg.Truncate(truncateSize)
+	}
+	err := w.WriteMsg(msg)
+	if err != nil {
```
mook-as: (Missed `if err := … ; err != nil…` opportunity.)

pkg/test/server.go
```diff
@@ -0,0 +1,125 @@
...
+		logrus.Errorf("ServeDNS trying to write message in udpHandler: %v", err)
+	}
+}
+
+func parseReply(msg *dns.Msg) {
+	logrus.Infof("ServeDNS handling request: %v", msg.Question)
```
mook-as: Do we expect to hammer this server a lot? If yes, `Info` may be too high a level for logging.

pkg/test/server.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, q := range msg.Question {
+		switch q.Qtype { //nolint:gocritic // this will have additional cases soon
+		case dns.TypeA:
+			msg.Authoritative = true
```
mook-as: Do we expect to be authoritative? I wouldn't expect it to be for the general case (I'd expect the upstream of user systems to be the DNS server at their ISP, which would typically not be authoritative for most of the internet.)

mook-as (changes requested):

pkg/test/server.go
Nino-K: This is not going to act as a recursive server, this will be the final upstream. There is no internet and we are planning to tap into win32 DNS API since it won't work without internet.

pkg/test/server.go
Nino-K: This will be the final upstream server and will not be acting as a recursive server since the plan is not to have any lookup using the net. Also, win32 API will not work without the internet.

mook-as (approved):

PR #20, opened 2022-05-13
Adds e2e tests

Nino-K:
TODO items:
- [x] the test runner needs to be elevated or run as admin (this could no longer be an issue since github self-hosted runners do run as admin)
- [ ] add all supported resource records other than A records


pkg/helper/net_windows.go
```diff
@@ -0,0 +1,43 @@
...
+limitations under the License.
+*/
+package helper
+
+import (
+	"golang.org/x/sys/windows"
+	"golang.zx2c4.com/wireguard/windows/tunnel/winipcfg"
+)
+
+// GetDnsInterfaces returns network interfaces that have active DNS servers associtatd to them.
```
mook-as:
```suggestion
// GetDnsInterfaces returns network interfaces that have active DNS servers associated to them.
```

pkg/helper/net_windows.go
```diff
@@ -0,0 +1,43 @@
...
+// The filtering criteria requests all address families using AF_UNSPEC since interfaces can have
+// DNS server addrs for IPv6 and not IPv4 vice versa. It further checks the operation status of the
```
mook-as:
```suggestion
// DNS server addrs for IPv6 and not IPv4, or vice versa. It further checks the operation status of the
```

pkg/helper/net_windows.go
```diff
@@ -0,0 +1,43 @@
...
+// adapter's address to make sure it is up. The final filter is to remove all the addresses without a
+// DCHP server, this filter eliminates interfaces that are not dynamically configured
+// e.g. VirtualBox Host-Only Network, vEthernet (WSL)
```
mook-as, line 26: What if I have a static IP address configured manually (which would of course still require a DNS server)?

pkg/helper/net_windows.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var activeIfs []*winipcfg.IPAdapterAddresses
+	for _, addr := range addrs {
+		dnsAddrs, err := addr.LUID.DNS()
+		if err != nil {
+			return nil, err
+		}
+		if addr.OperStatus == winipcfg.IfOperStatusUp && len(dnsAddrs) != 0 && addr.DHCPv4Server.Sockaddr != nil {
```
mook-as, line 38: - Why are we filtering out IPv4 with static IP?
- Why are we filtering out IPv6-only (i.e. no IPv4 DHCP server)?

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+	"net"
+	"sort"
+	"time"
+
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/test/testdns"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// 10s timeout should be adequate
```
mook-as: Please make the comment to the right of the line instead (so it doesn't show up as a doc comment).

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		return dnsQuery(addr, n, records)
+	},
+}
+
+func init() {
+	dnshammerCmd.Flags().StringP("server-address", "a", "127.0.0.1:53", "Address of the DNS server.")
+	dnshammerCmd.Flags().StringToStringP("rr-type", "r", map[string]string{},
+		`List of desired resource records mapped to the csv test data file.
+	Supported records are: A, CNAME, TXT, NS, MX, SRV. Accepted Format: A=Arecords.csv,CNAME=cnames.csv`)
```
mook-as: We actually only support `A`?

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+	rootCmd.AddCommand(dnshammerCmd)
+}
+
+func dnsQuery(srvAddr string, n int, records map[string]string) error {
+	for rr, path := range records {
+		switch rr { //nolint:gocritic // this will have additional cases soon
+		case "A":
+			arecords := testdns.LoadRecords(path)
+			err := do(srvAddr, n, arecords)
+			if err != nil {
```
mook-as:
```suggestion
			if err := do(srvAddr, n, arecords); err != nil {
```

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+		return dnsQuery(addr, n, records)
+	},
+}
+
+func init() {
+	dnshammerCmd.Flags().StringP("server-address", "a", "127.0.0.1:53", "Address of the DNS server.")
+	dnshammerCmd.Flags().StringToStringP("rr-type", "r", map[string]string{},
+		`List of desired resource records mapped to the csv test data file.
+	Supported records are: A, CNAME, TXT, NS, MX, SRV. Accepted Format: A=Arecords.csv,CNAME=cnames.csv`)
+	dnshammerCmd.Flags().IntP("request-number", "n", defaultRequestNumber, "Number of request against the DNS server.")
```
mook-as: `0` (i.e. the default) appears to have special meaning of some sort?

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func dnsLookup(addr, resolverProtocol, domain string) ([]net.IP, error) {
+	resolver := net.Resolver{
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			dialer := net.Dialer{}
+			return dialer.DialContext(ctx, resolverProtocol, addr)
+		},
+	}
+	// maybe change to debug
```
mook-as: It's already debug?

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func dnsLookup(addr, resolverProtocol, domain string) ([]net.IP, error) {
+	resolver := net.Resolver{
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			dialer := net.Dialer{}
```
mook-as, line 121: Should we reuse the dialer?

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,132 @@
...
+			return dialer.DialContext(ctx, resolverProtocol, addr)
+		},
+	}
+	// maybe change to debug
+	logrus.Debugf("[DNS] lookup on %s [%s] -> %s", addr, resolverProtocol, domain)
+	ctx, cancel := context.WithTimeout(context.Background(), lookupTimeoutSec*time.Second)
```
mook-as: Shouldn't a context get passed in? (We'd still need to set a new timeout per lookup, of course.)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
+	"golang.zx2c4.com/wireguard/windows/tunnel/winipcfg"
+)
+
+var (
+	// for now we use this maybe this can be an env var from test
+	wslDistroName = "Ubuntu"
+	testSrvAddr   = "127.0.0.1"
+	dnsPort       = "53"
```
mook-as: Yeah, that probably shouldn't be fixed

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+func TestLookupARecords(t *testing.T) { //nolint:funlen
```
mook-as: In general I think function length lints are dumb.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	tmpDir, err := ioutil.TempDir("", "e2e_test_")
```
mook-as: Why not just use [`t.TempDir()`](https://pkg.go.dev/testing@go1.18.2#T.TempDir)?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	require.NoError(t, err, "Failed creating temp dir")
+	defer os.RemoveAll(tmpDir)
+
+	t.Logf("Creating %v wsl distro", wslDistroName)
+	installCmd := cmdExec("wsl", "--user", "root",
+		"--install",
+		"--distribution", wslDistroName)
```
mook-as: Hmm, doesn't `wsl --install` end up spawning a new process (with a detached console for the first-run stuff)?

Can we get away with `wsl --import` (which just extracts a tarball, and therefore has no first-run)? It would mean having something to build the tarball though.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/helper" //nolint:ignore
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/test/testdns"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
+	"golang.zx2c4.com/wireguard/windows/tunnel/winipcfg"
+)
+
+var (
+	// for now we use this maybe this can be an env var from test
+	wslDistroName = "Ubuntu"
```
mook-as: Given that we dispose of the distro afterwards, calling it `Ubuntu` seems not awesome.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}()
+
+	// It takes a long time to start a new distro,
+	// 20 sec is a long time but that's actually how long
+	// it takes to start a distro without any flakeyness
+	timeout := time.Second * 20
+	tryInterval := time.Second * 2
+	err = confirm(func() bool {
+		// this is a way to figure out if the distro is running
+		// there is an issue with wsl --list --running output
```
mook-as: Having a comment on _what_ the issue is would be great.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+		out, err := cmdRunWithOutput("wsl", "--distribution", wslDistroName, "--exec", "echo", wslDistroName)
```
mook-as: It might be a good idea to run `/bin/wslpath .` instead; `echo` requires that there's an echo executable (or a shell), whereas `/bin/wslpath` is provided by WSL itself (so we have no dependencies on what's in the distro).

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	// This is to cache all the exsisting DNS addresses
+	guidToDNSAddr, err := cacheExsitingDNSAddrs(dnsInfs)
+	require.NoError(t, err, "Failed caching exsisting DNS server addresses")
+
+	t.Log("Updating network interfaces with test DNS server addr")
+	// Update the dns addrs to test server
+	updateSystemDNS(t, dnsInfs)
+
+	// restore the system DNS servers to the original state
+	defer restoreSystemDNS(t, dnsInfs, guidToDNSAddr)
```
mook-as: Might want to set up the restore _before_ we do the update, in case `updateSystemDNS` panics.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+		"--exec", "./test", "dnshammer",
+		"--server-address", dnsSrvAddr,
+		"--rr-type", "A=test-300.csv")
+	runTestCmd.Dir = tmpDir
+	err = runTestCmd.Run()
+	require.NoError(t, err, "Running dns hammer against the peer process faild")
+
+	_ = hostCmd.Process.Kill()
+	_ = peerCmd.Process.Kill()
+	_ = runTestCmd.Process.Kill()
```
mook-as: These should be set up as `defer`s right after those processes get set up. (It's also possible to use a context instead.)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// TODO (Nino-K): maybe this can be replaced by CI
+func buildBinaries(path, goos, tmpDir string) error {
+	buildCmd := exec.Command("go", "build", "-o", tmpDir, path)
+	buildCmd.Env = os.Environ()
+	buildCmd.Env = append(buildCmd.Env, fmt.Sprintf("GOOS=%s", goos), "GOARCH=amd64")
```
mook-as: I don't think we really need `GOARCH`?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	err := cmd.Run()
+	if err != nil {
+		return errBuf.String(), err
+	}
+	return outBuf.String(), nil
+}
+
+func cmdExec(command string, args ...string) *exec.Cmd {
```
mook-as: Most callers seem to also want to set `.Dir`; should that be built in?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	bytesRead, err := os.ReadFile(src)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(dst, bytesRead, 0600)
+}
+
+func cacheExsitingDNSAddrs(adapterAddrs []*winipcfg.IPAdapterAddresses) (map[string][]netip.Addr, error) {
+	var adapterAddrsCopy = make([]*winipcfg.IPAdapterAddresses, len(adapterAddrs))
+	copy(adapterAddrsCopy, adapterAddrs)
```
mook-as: Huh, does this copy the pointer or the contents? (Given that both src and dest are slices of pointers)

mook-as: Actually, I'm not even sure what the point of copy was; we never modify `adapterAddrs`, and `adapterAddrsCopy` goes out of scope when the function returns…

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+}
+
+func updateSystemDNS(t *testing.T, dnsInfs []*winipcfg.IPAdapterAddresses) {
+	testDNSAddr := netip.MustParseAddr(testSrvAddr)
+	testDNSAddrIPv6 := netip.IPv6Unspecified()
+	for _, addr := range dnsInfs {
+		// Set IPv4 DNS
+		err := addr.LUID.SetDNS(winipcfg.AddressFamily(windows.AF_INET), []netip.Addr{testDNSAddr}, []string{})
```
mook-as: Why does this require a `AddressFamily` cast, but `AF_INET6` doesn't?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+
+func restoreSystemDNS(t *testing.T, addrs []*winipcfg.IPAdapterAddresses, cachedAddrs map[string][]netip.Addr) {
+	t.Log("Restoring DNS servers back to the original state")
+	for _, addr := range addrs {
+		addrGUID, err := addr.LUID.GUID()
+		require.NoErrorf(t, err, "Failed getting interface GUID for: %v", addr.FriendlyName())
+		if dnsAddr, ok := cachedAddrs[addrGUID.String()]; ok {
+			err := addr.LUID.SetDNS(windows.AF_INET, dnsAddr, []string{})
```
mook-as:
```suggestion
			err = addr.LUID.SetDNS(windows.AF_INET, dnsAddr, []string{})
```
We don't need to shadow the outer `err`.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+			require.NoErrorf(t, err, "Failed setting DNS server IPv4 addrs for: %v", addr.FriendlyName())
+			err = addr.LUID.SetDNS(windows.AF_INET6, dnsAddr, []string{})
```
mook-as: We're always wiping out the search domains here; that seems wrong?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+func updateSystemDNS(t *testing.T, dnsInfs []*winipcfg.IPAdapterAddresses) {
+	testDNSAddr := netip.MustParseAddr(testSrvAddr)
+	testDNSAddrIPv6 := netip.IPv6Unspecified()
+	for _, addr := range dnsInfs {
+		// Set IPv4 DNS
+		err := addr.LUID.SetDNS(winipcfg.AddressFamily(windows.AF_INET), []netip.Addr{testDNSAddr}, []string{})
+		require.NoErrorf(t, err, "Failed setting IPv4 DNS server for: %v", addr.FriendlyName())
+		// Set IPv6 DNS to unspecified so DNS lookup will not be bypassed
+		err = addr.LUID.SetDNS(windows.AF_INET6, []netip.Addr{testDNSAddrIPv6}, []string{})
```
mook-as, line 256: Would using [`FlushDNS`](https://pkg.go.dev/golang.zx2c4.com/wireguard/windows/tunnel/winipcfg#LUID.FlushDNS) make more sense here?

mook-as (changes requested):
`test/LICENSE` appears to be empty; please either remove it or fill it with a license.

Should add instructions on how to run the tests.

pkg/helper/net_windows.go
Nino-K: We don't have that use case yet, this assumes the CI servers machines will be DHCP however if it turns out to be an issue we can add extra stuff here.

pkg/helper/net_windows.go
Nino-K: > * Why are we filtering out IPv4 with static IP?

I'm trying to avoid interfaces like VirtualBox Host-Only Network, and vEthernet (WSL) on the CI machine, since updating the DNS on those interfaces could be problematic and causes disconnection during the test. 

> * Why are we filtering out IPv6-only (i.e. no IPv4 DHCP server)?

There is a high chance that IPv6 DHCP will be nil.

test/cmd/dnshammer.go
Nino-K: No, this test only implements `A` for now, I will be adding other RRs soon.

test/cmd/dnshammer.go
Nino-K: I will, once I add the other types I can get a better idea, of what it should look like.

test/cmd/dnshammer.go
Nino-K: Technically it should, but not very useful in our use case.

test/e2e/dnslookup_test.go
Nino-K: Yeah, the popup window is annoying and if we are considering disposing of the VM after since this will run on a github self-hosted runner, It may not be super bad. I considered the tarball, I just don't have a clear picture right now on how to produce the artifact, maybe we can visit this once we are setting up the CI.

test/e2e/dnslookup_test.go
Nino-K: Yes, I agree. It no longer makes sense, got left behind from refactoring.

Nino-K:
@mook-as the test need to run in an elevated shell as an admin by running `go test -v .\test\e2e\dnslookup_test.go`

pkg/helper/net_windows.go
mook-as: Then this needs to be called out in a readme, because the failure will be _very_ confusing.

test/e2e/dnslookup_test.go
mook-as: Do we have any particular needs from the distro? Because we _do_ have https://github.com/rancher-sandbox/rancher-desktop-wsl-distro just sitting there…

mook-as (changes requested):
Please provide test instructions in a readme file or something, so that people don't need to dig into a PR to find the instructions.

(I recognize that we'll want to make this run on CI; but until that lands, people still need a way to test things for other PRs.)

README.md
```diff
@@ -53,5 +53,18 @@ You can run the tests in the container by running:
 ```bash
 docker build -t host-resolver:latest . && docker run --dns 127.0.0.1 -it host-resolver:latest
 ```
+## E2E Test
+
+You can run the e2e tests locally on a windows machine, please note that the e2e tests need to run
+as an administrator in an elevated terminal (e.g. Administrator: Windows Powershell).
```
mook-as:
```suggestion
as an administrator in an elevated terminal (e.g. Administrator: Windows PowerShell).
```

README.md
```diff
@@ -53,5 +53,18 @@ You can run the tests in the container by running:
...
+
+```bash
```
mook-as:
```suggestion
```pwsh
```

README.md
```diff
@@ -53,5 +53,18 @@ You can run the tests in the container by running:
...
+go test -v .\test\e2e\...
+```
+NOTE: the e2e test updates the DNS addresses on the machine's PRIMARY interface (e.g WiFi, eth0, etc)
```
mook-as:
```suggestion
NOTE: the e2e test updates the DNS addresses on the machine's **primary** interface (e.g WiFi, eth0, etc.).
```

README.md
```diff
@@ -53,5 +53,18 @@ You can run the tests in the container by running:
...
+the determination process assumes the addresses are dynamically configured through DHCP, this is to prevent
```
mook-as:
```suggestion
The determination process assumes the addresses are dynamically configured through DHCP, this is to prevent
```

pkg/helper/net_windows.go
mook-as: I still think this is wrong, but it's probably close enough to right that we can accept it and fix it later.

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,130 @@
...
+	"net"
+	"sort"
+	"time"
+
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/test/testdns"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const lookupTimeoutSec = 10 // 10s timeout should be adequate
```
mook-as: I think this would be clearer expressed as:
```suggestion
const lookupTimeout = 10 * time.Second
```
(Also, the comment doesn't really say anything useful.)

test/cmd/dnshammer.go
mook-as: Right; I'm saying we should only list `A` here, and add the others as we implement them.

test/cmd/dnshammer.go
```diff
@@ -0,0 +1,130 @@
...
+	sort.Strings(b)
+	for i, ip := range a {
+		if ip != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func dnsLookup(addr, resolverProtocol, domain string) ([]net.IP, error) {
```
mook-as: Perhaps the function name should indicate that this is for `A` records?

test/e2e/dnslookup_test.go
mook-as: Actually, #22 killed this, so we shouldn't need it anymore.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	t.Log("Restoring DNS servers back to the original state")
+	for _, addr := range addrs {
+		err := addr.LUID.FlushDNS(windows.AF_INET)
+		require.NoErrorf(t, err, "Failed to flush DNS for IPv4 addrs for: %v", addr.FriendlyName())
+		err = addr.LUID.FlushDNS(windows.AF_INET6)
+		require.NoErrorf(t, err, "Failed to flush DNS for IPv6 addrs for: %v", addr.FriendlyName())
+	}
+}
+
+func downloadWSLTarball(path, url string) error {
```
mook-as: Nothing in this function actually cares that this is a tarball.
```suggestion
func downloadFile(path, url string) error {
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+func TestLookupARecords(t *testing.T) { //nolint:funlen
+	tmpDir := t.TempDir()
+
+	t.Logf("Dowloading %v wsl distro tarball", wslTarballName)
+	tarballPath := filepath.Join(tmpDir, wslTarballName)
+
+	err := downloadWSLTarball(tarballPath, wslTarballURL)
+	require.NoErrorf(t, err, "Failed to download wsl distro tarball %v", wslTarballName)
+
+	t.Logf("Creating %v wsl distro", wslDistroName)
```
mook-as, line 74: We should eventually extract this out, since we'd need it for other tests too. But that doesn't need to be in this PR.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+		wslDistroName,
+		".",
+		tarballPath)
+	err = installCmd.Run()
+	require.NoErrorf(t, err, "Failed to install distro %v", wslDistroName)
+
+	defer func() {
+		t.Logf("Deleting %v wsl distro", wslDistroName)
+		_, err = cmdRunWithOutput("wsl", "--unregister", wslDistroName)
+		require.NoErrorf(t, err, "Failed to start distro %v", wslDistroName)
```
mook-as:
```suggestion
		require.NoErrorf(t, err, "Failed to unregister distro %v", wslDistroName)
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	}()
+
+	// It takes a long time to start a new distro,
+	// 20 sec is a long time but that's actually how long
+	// it takes to start a distro without any flakeyness
```
mook-as:
```suggestion
	// it takes to start a distro without any flakiness
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	timeout := time.Second * 20
```
mook-as, line 95:
```suggestion
	timeout := 20 * time.Second
```
(It reads better as English)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	tryInterval := time.Second * 2
+	err = confirm(func() bool {
+		// this is a way to figure out if the distro is running
+		// there is an issue with wsl --list --running output
+		// the buffer is not very useful for string search since it
+		// returns some sort of unicode.
```
mook-as: UTF16-LE, in particular.  That's really the internal Windows encoding.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+		out, err := cmdRunWithOutput("wsl", "--distribution", wslDistroName, "--exec", "/bin/wslpath", ".")
+		if err != nil {
+			return false
+		}
+		// remove all the weirdness from WSL output
```
mook-as: Which weirdness? Could you explain this better?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+func cmdRunWithOutput(command string, args ...string) (string, error) {
+	var outBuf, errBuf strings.Builder
+	cmd := exec.Command(command, args...)
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	err := cmd.Run()
+	if err != nil {
+		return errBuf.String(), err
+	}
+	return outBuf.String(), nil
```
mook-as, line 211: Returning a string is incorrect here, since it's stuff UTF16LE encoded (in the case of `wsl.exe`).

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+		// this is a way to figure out if the distro is running
+		// there is an issue with wsl --list --running output
+		// the buffer is not very useful for string search since it
+		// returns some sort of unicode.
+		out, err := cmdRunWithOutput("wsl", "--distribution", wslDistroName, "--exec", "/bin/wslpath", ".")
+		if err != nil {
+			return false
+		}
+		// remove all the weirdness from WSL output
+		return strings.TrimSpace(out) == "."
```
mook-as, line 105: We should have a better comment on what it's doing (specifically, that `wslpath .` should return a literal `.` as response).

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	}, tryInterval, timeout)
+	require.NoErrorf(t, err, "Failed to check if %v wsl distro is running", wslDistroName)
+
+	dnsInfs, err := helper.GetDNSInterfaces()
+	require.NoError(t, err, "Failed getting DNS addrs associated to interfaces")
+
+	// This is to cache all the exsisting DNS addresses
```
mook-as: It's not really a cache; rather, it's the original settings that we're about to clobber.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	guidToDNSAddr, err := cacheExsitingDNSAddrs(dnsInfs)
```
mook-as:
```suggestion
	guidToDNSAddr, err := getExistingDNSAddrs(dnsInfs)
```
(This includes a typo in _existing_)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	require.NoError(t, err, "Failed caching exsisting DNS server addresses")
+
+	// restore the system DNS servers to the original state
+	defer restoreSystemDNS(t, dnsInfs, guidToDNSAddr)
+
+	t.Log("Updating network interfaces with test DNS server addr")
```
mook-as:
```suggestion
	t.Log("Updating network interfaces with test DNS server address")
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+				return nil
+			}
+		case <-terminate:
+			return fmt.Errorf("Failed to run within %v", timeout)
+		}
+	}
+}
+
+func updateSystemDNS(t *testing.T, dnsInfs []*winipcfg.IPAdapterAddresses) {
+	testDNSAddr := netip.MustParseAddr(testSrvAddr)
```
mook-as, line 249: Can we pass `testSrvAddr` in? I know it's a global constant, but limiting the use of the global still seems like a good idea.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	require.NoError(t, err, "Failed caching exsisting DNS server addresses")
+
+	// restore the system DNS servers to the original state
+	defer restoreSystemDNS(t, dnsInfs, guidToDNSAddr)
+
+	t.Log("Updating network interfaces with test DNS server addr")
+	// Update the dns addrs to test server
+	updateSystemDNS(t, dnsInfs)
+
+	t.Log("Building host-resolver host binary")
```
mook-as: Wouldn't it make more sense to build things before playing with DNS? That way if the build fails we wouldn't need to worry about fixing DNS.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	require.NoError(t, err, "Failed building host-resolver.exe")
+
+	t.Log("Building host-resolver peer binary")
+	err = buildBinaries("../../...", "linux", tmpDir)
+	require.NoError(t, err, "Failed building host-resolver")
+
+	aRecords := testdns.LoadRecords("../testdata/test-300.csv")
+
+	tcpHandler := testdns.NewHandler(false)
+	tcpHandler.Arecords = aRecords
```
mook-as: Hmm, maybe we could instead do:
```suggestion
	tcpHandler := &testdns.NewHandler{
		Truncate: false,
		Arecords: aRecords,
	}
```
(Of course, that means exporting `Truncate`; not sure we want that…)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	defer func() {
+		_ = hostCmd.Process.Kill()
+	}()
+
+	t.Log("Building DNS hammer binary")
+	err = buildBinaries("../...", "linux", tmpDir)
+	require.NoError(t, err, "Failed building dnsHammer")
+
+	err = copyTestData("../testdata/test-300.csv", fmt.Sprintf("%s/test-300.csv", tmpDir))
+	require.NoError(t, err, "Failed copying test data file")
```
mook-as: We should build binaries before running servers.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	t.Log("Confirming host-resolver peer process is up")
+	peerCmdTimeout := time.Second * 10
+	confirm(func() bool { //nolint:errcheck // we don't care about the error
+		p, _ := os.FindProcess(peerCmd.Process.Pid)
```
mook-as: If there was an error (maybe the pid doesn't exist), then `p` would be `nil`, so accessing `p.Pid` on the next line would panic?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+		return p.Pid == peerCmd.Process.Pid
+	}, tryInterval, peerCmdTimeout)
+
+	t.Logf("Running dns hammer test process in wsl [%v]", wslDistroName)
```
mook-as:
```suggestion
	t.Logf("Running DNS hammer test process in WSL distribution [%v]", wslDistroName)
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	err = runTestCmd.Run()
+	require.NoError(t, err, "Running dns hammer against the peer process faild")
+	_ = runTestCmd.Process.Kill()
+}
+
+// TODO (Nino-K): maybe this can be replaced by CI
+func buildBinaries(path, goos, tmpDir string) error {
+	buildCmd := exec.Command("go", "build", "-o", tmpDir, path)
+	buildCmd.Env = os.Environ()
+	buildCmd.Env = append(buildCmd.Env, fmt.Sprintf("GOOS=%s", goos))
```
mook-as:
```suggestion
	buildCmd.Env = append(os.Environ(), fmt.Sprintf("GOOS=%s", goos))
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+	cmd := exec.Command(command, args...)
+	if execDir != "" {
+		cmd.Dir = execDir
+	}
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd
+}
+
+func copyTestData(src, dst string) error {
```
mook-as:
```suggestion
func copyFile(src, dst string) error {
```
This doesn't rely on the file being test data.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+			return nil, err
+		}
+		guidToDNSAddrs[guid.String()] = dnsAddrs
+	}
+	return guidToDNSAddrs, nil
+}
+
+func confirm(command func() bool, interval, timeout time.Duration) error {
+	tick := time.NewTicker(interval)
+	terminate := time.After(timeout)
```
mook-as, line 234: Semantically, this is basically `context.WithTimeout`, right?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,299 @@
...
+		// Set IPv4 DNS
+		err := addr.LUID.SetDNS(windows.AF_INET, []netip.Addr{testDNSAddr}, []string{})
+		require.NoErrorf(t, err, "Failed setting IPv4 DNS server for: %v", addr.FriendlyName())
+		// Set IPv6 DNS to unspecified so DNS lookup will not be bypassed
+		err = addr.LUID.SetDNS(windows.AF_INET6, []netip.Addr{testDNSAddrIPv6}, []string{})
+		require.NoErrorf(t, err, "Failed setting IPv6 DNS server for: %v", addr.FriendlyName())
+	}
+}
+
+func restoreSystemDNS(t *testing.T, addrs []*winipcfg.IPAdapterAddresses, cachedAddrs map[string][]netip.Addr) {
```
mook-as: This doesn't appear to use `cachedAddrs` anymore; are we actually restoring things?
(I _do_ notice that it seems to work; I'm not sure why, though.)

mook-as (changes requested):
The tests seem to pass for me (on a VM).

.golangci.yml
```diff
@@ -60,7 +60,6 @@ linters:
     - dupl
     - errcheck
     - exportloopref
-    - funlen
```
mook-as: Wasn't this covered by #22? Do you need to rebase?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"net/http"
+	"net/netip"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/helper" //nolint:ignore
```
mook-as: Why does this need nolint?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,285 @@
...
+		Truncate: true,
+		Arecords: aRecords,
+	}
+
+	testServer := testdns.Server{
+		Addr:       testSrvAddr,
+		TCPPort:    dnsPort,
+		UDPPort:    dnsPort,
+		TCPHandler: tcpHandler,
+		UDPHandler: udpHandler,
```
mook-as, line 136: Hmm, these can be inlined now (up to you).

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,285 @@
...
+	t.Log("Building DNS hammer binary")
+	err = buildBinaries("../...", "linux", tmpDir)
+	require.NoError(t, err, "Failed building dnsHammer")
+
+	err = copyFile("../testdata/test-300.csv", fmt.Sprintf("%s/test-300.csv", tmpDir))
+	require.NoError(t, err, "Failed copying test data file")
+
+	t.Log("Confirming host-resolver peer process is up")
+	peerCmdTimeout := time.Second * 10
+	confirm(func() bool { //nolint:errcheck
```
mook-as: Why are we ignoring the error from `confirm`? At least do `_ = confirm(…` instead of suppressing the linter?

test/e2e/dnslookup_test.go
mook-as: (This is obsolete — we always expect UTF8 here now)

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,285 @@
...
+		"--import",
+		wslDistroName,
+		".",
+		tarballPath)
+	err = installCmd.Run()
+	require.NoErrorf(t, err, "Failed to install distro %v", wslDistroName)
+
+	defer func() {
+		t.Logf("Deleting %v wsl distro", wslDistroName)
+		_, err = cmdRunWithOutput("wsl", "--unregister", wslDistroName)
```
mook-as: Why are we not using `cmdExec` here, if we don't want the output?

mook-as (changes requested):
There's also a couple comments from previous reviews that could use some reply.

test/cmd/root.go
```diff
@@ -0,0 +1,36 @@
...
+
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
```
mook-as: Just noticed that this was here (and not doing much).  It's harmless enough that I'm okay with dropping it in a follow-up.

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,287 @@
...
+		"--distribution", wslDistroName,
+		"--exec", "./rancher-desktop-host-resolver", "vsock-peer")
+	err = peerCmd.Start()
+	require.NoError(t, err, "Starting host-resolver peer process faild")
+	defer func() {
+		_ = peerCmd.Process.Kill()
+	}()
+
+	t.Log("Starting host-resolver host process")
+	resolverExecPath := fmt.Sprintf("%v/rancher-desktop-host-resolver.exe", tmpDir)
```
mook-as: Shouldn't this be `path.Join`?

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,287 @@
...
+	hostCmd := cmdExec(
+		tmpDir,
+		resolverExecPath, "vsock-host",
+		"--upstream-servers", fmt.Sprintf("[%v]", testSrvAddr))
+	err = hostCmd.Start()
+	require.NoError(t, err, "Starting host-resolver host process faild")
+	defer func() {
+		_ = hostCmd.Process.Kill()
+	}()
```
mook-as, line 163: (Optional)
```suggestion
	defer require.NoError(t, hostCmd.Process.Kill(), "Failed to stop vsock-host process")
```

test/e2e/dnslookup_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+	t.Log("Building DNS hammer binary")
+	err = buildBinaries("../...", "linux", tmpDir)
+	require.NoError(t, err, "Failed building dnsHammer")
+
+	err = copyFile("../testdata/test-300.csv", fmt.Sprintf("%s/test-300.csv", tmpDir))
+	require.NoError(t, err, "Failed copying test data file")
```
mook-as: We're still building binaries and copying files after starting up processes? Huh.

mook-as (approved):

mook-as (approved):

PR #23, opened 2022-06-06
Fixes DNS truncate for responses longer than 512 bytes issue

### LINKED ISSUE #17: Investigate why host-resolver eats up responses longer than 512 bytes
Request for resource records with longer replies (specifically TXT records) seems to be broken. The correct behavior for the host-resolver would be to upgrade to TCP for any response longer than `512`. 

Also, direct request for `TXT` records over `TCP` doesn't seem to work.

```
 dig +tcp -t txt suse.com
```

However, the underlying stub DNS truncates everything down to 512 bytes [here](https://github.com/rancher-sandbox/rancher-desktop-host-resolver/blob/325fb3dc92af0804f94c8e2c0da3d5e6f195ba21/pkg/dns/dns.go#L252) as a workaround, we will be separating UDP/TCP handlers to do the truncation accordingly. 

This issue was discovered while testing the following issue:
https://github.com/rancher-sandbox/rancher-desktop/issues/2141

Nino-K:
Fixes https://github.com/rancher-sandbox/rancher-desktop-host-resolver/issues/17


pkg/vmsock/conn_windows.go
```diff
@@ -22,12 +22,12 @@ import (
...
 )
 
-func Listen() (net.Listener, error) {
-	vmGUID, err := vmGUID()
-	if err != nil {
-		return nil, fmt.Errorf("Listen, could not determine VM GUID: %v", err)
-	}
-	svcPort, err := hvsock.GUIDFromString(winio.VsockServiceID(HostListenPort).String())
+func GetVMGUID() (hvsock.GUID, error) {
+	return vmGUID()
```
mook-as: Why not just rename and export `vmGUID` directly?

Also, exported functions should get doc comments.

pkg/commands/start_standalone.go
```diff
@@ -77,7 +77,7 @@ func randomUDPPort() (int, error) {
 	return l.LocalAddr().(*net.UDPAddr).Port, nil
 }
 
-func run() {
+func trapSignal() {
```
mook-as: Consider naming this something more obvious, like `waitForExitSignal`

pkg/dns/dns.go
```diff
@@ -67,7 +69,7 @@ func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
 	return dns.ClientConfigFromReader(r)
 }
 
-func newHandler(IPv6 bool, hosts map[string]string, upstreamServers []string) (dns.Handler, error) {
+func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []string) (dns.Handler, error) {
```
mook-as, line 72: There's so many things passing in here now I wonder if just passing in `ServerOptions` would be a good idea.

If you want to make it nicer, make the subset into a `HandlerOptions` that's embedded into `ServerOptions`.

test/cmd/dnshammer.go
```diff
@@ -94,11 +102,34 @@ func do(addr string, n int, records map[string][]string) error {
 	return nil
 }
 
-func ipToString(ips []net.IP) (out []string) {
-	for _, ip := range ips {
-		out = append(out, ip.String())
+func doTXT(addr string, n int, records map[string][]string) error {
```
mook-as: This is almost the same as `doA`; can we merge the two (and passing in a lookup function returning `[]string, error`, plus a compare function of type `func([]string, []string) bool`)?

test/cmd/dnshammer.go
```diff
@@ -115,6 +146,13 @@ func compare(a, b []string) bool {
 	return true
 }
 
+func ipToString(ips []net.IP) (out []string) {
```
mook-as: We can drop this if `lookupARecord` returns a `[]string` instead.

test/cmd/dnshammer.go
```diff
@@ -127,3 +165,44 @@ func lookupARecord(addr, resolverProtocol, domain string) ([]net.IP, error) {
...
 	return resolver.LookupIP(ctx, "ip4", domain)
 }
+
+func lookupTXT(domain string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), lookupTimeout)
+	defer cancel()
+	return net.DefaultResolver.LookupTXT(ctx, domain)
+}
+
+func loadRecords(p string) map[string][]string {
```
mook-as, line 132: Why do we load them from disk, instead of just generating them here (and never hitting disk)?

mook-as (changes requested):

pkg/dns/dns.go
Nino-K: I have a cleanup PR coming up just for `dns.go` file. Let's revisit in that PR.

test/cmd/dnshammer.go
Nino-K: There is a disconnect between the test runner and the `dnsHammer` which is a standalone binary. The test generates the data in memory and writes it to the file for the `dnsHammer` to consume. Also, the records needs to be shared between the test DNS server and the `dnsHammer` so generating the records in the `dnsHammer` adds more complexity.

test/cmd/dnshammer.go
Nino-K: I considered merging these, however, the lookup functions have a different argument signature and in addition, we would need to add more lookup funcs in the close future, e.g lookupMX, lookupSRV, lookupHost, etc. They will all have a diffrent signature, if we refactor now, we would need to decouple this shortly.

test/cmd/dnshammer.go
mook-as: Right, but _why_ does the test generate the test data, instead of letting `dnsHammer` do it?

mook-as: … because it needs in both the server and dnshammer.

mook-as (changes requested):
Per offline conversation, Nino is going to normalize `lookupARecord` / `lookupTXT` to both use a custom resolver, which will be passed in to the function.

pkg/commands/start_host_windows.go
```diff
@@ -19,24 +19,50 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// StartVsockHost attempts to start two AF_VSOCK listeners, one acting
+// for TCP (stream) and the other for UDP(datagram) it then waits for the
```
mook-as:
```suggestion
// for TCP (stream) and the other for UDP (datagram); it then waits for the
```

test/cmd/dnshammer.go
```diff
@@ -56,36 +56,40 @@ func init() {
...
+			if err := verifyResults(net.LookupTXT, contains, n, srvAddr, path); err != nil {
 				return err
 			}
 		}
 	}
 	return nil
 }
 
-func do(addr string, n int, records map[string][]string) error {
+func verifyResults(lookup func(string) ([]string, error), assert func([]string, []string) bool, n int, addr, path string) error {
```
mook-as: `addr` doesn't seem to be used anymore.

mook-as (changes requested):

test/cmd/dnshammer.go
```diff
@@ -48,57 +44,64 @@ var dnshammerCmd = &cobra.Command{
...
 	}
-	logrus.Infof("Successfully tested %v records, against: %v", i, addr)
+	logrus.Infof("Successfully tested %v records", i)
 	return nil
 }
 
-func ipToString(ips []net.IP) (out []string) {
-	for _, ip := range ips {
-		out = append(out, ip.String())
+func contains(expected, actual []string) bool {
```
mook-as, line 99: So this tests that, for each string in `expected`, it's a substring of something in `actual`? That is, `contains([]string{"hello", "world"}, []string{"hello world", "completely unused"})` should pass?

That's… fine, but unexpected. Might want to document that.

test/cmd/dnshammer.go
```diff
@@ -115,15 +118,45 @@ func compare(a, b []string) bool {
...
+	for _, line := range fileData {
+		var data = []string{}
+		for j, field := range line {
+			if j == 0 {
+				domain = field
+			} else {
+				data = append(data, field)
+			}
+			records[domain] = data
+		}
```
mook-as: I just realized… isn't this the same as:
```suggestion
		if len(line) > 0 {
			records[line[0]] = line[1:]
		}
```

mook-as (changes requested):
Just need some clarification in `contains`.

test/cmd/dnshammer.go
Nino-K: Well, it's more like `contains([]string{"hello", "world"}, []string{"helloworld"})`... this is due to the way go TXT lookup resolver concatenates the strings.

mook-as (approved):

PR #24, opened 2022-06-15
Prepare dns.go for upstream

Nino-K:
This is a preliminary PR for changes that are required in the upstream `dns.go`.

### DIFF
--- a/pkg/commands/start_host_windows.go
+++ b/pkg/commands/start_host_windows.go
@@ -16,9 +16,10 @@ package commands
 import (
 	"github.com/linuxkit/virtsock/pkg/hvsock"
 	"github.com/miekg/dns"
+	"github.com/sirupsen/logrus"
+
 	rddns "github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
 	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/vmsock"
-	log "github.com/sirupsen/logrus"
 )
 
 // StartVsockHost attempts to start two AF_VSOCK listeners, one acting
@@ -30,7 +31,7 @@ func StartVsockHost(ipv6 bool, hosts map[string]string, upstreamServers []string
 		return err
 	}
 
-	tcpOpts := &rddns.ServerOptions{
+	tcpOpts := rddns.HandlerOptions{
 		IPv6:            ipv6,
 		StaticHosts:     hosts,
 		UpstreamServers: upstreamServers,
@@ -38,17 +39,17 @@ func StartVsockHost(ipv6 bool, hosts map[string]string, upstreamServers []string
 	}
 	tcpServer, err := startDNSServer(vmGUID, vmsock.HostTCPListenPort, tcpOpts)
 	if err != nil {
-		log.Panicf("StartVsockHost failed starting a TCP server: %v", err)
+		logrus.Panicf("StartVsockHost failed to start TCP DNS server: %v", err)
 	}
 
-	log.Infof("Started vsock-host AF_VSOCK stream server on VM: %v listening on port: %v", vmGUID.String(), vmsock.HostTCPListenPort)
+	logrus.Infof("Started vsock-host AF_VSOCK stream server on VM: %v listening on port: %v", vmGUID.String(), vmsock.HostTCPListenPort)
 	defer func() {
 		if err := tcpServer.Shutdown(); err != nil {
-			log.Errorf("Shutting down TCP server failed: %v", err)
+			logrus.Errorf("Shutting down TCP server failed: %v", err)
 		}
 	}()
 
-	udpOpts := &rddns.ServerOptions{
+	udpOpts := rddns.HandlerOptions{
 		IPv6:            ipv6,
 		StaticHosts:     hosts,
 		UpstreamServers: upstreamServers,
@@ -56,12 +57,12 @@ func StartVsockHost(ipv6 bool, hosts map[string]string, upstreamServers []string
 	}
 	udpServer, err := startDNSServer(vmGUID, vmsock.HostUDPListenPort, udpOpts)
 	if err != nil {
-		log.Panicf("StartVsockHost failed starting a UDP server: %v", err)
+		logrus.Panicf("StartVsockHost failed to start UDP DNS server: %v", err)
 	}
-	log.Infof("Started vsock-host AF_VSOCK datagram server on VM: %v listening on port: %v", vmGUID.String(), vmsock.HostUDPListenPort)
+	logrus.Infof("Started vsock-host AF_VSOCK datagram server on VM: %v listening on port: %v", vmGUID.String(), vmsock.HostUDPListenPort)
 	defer func() {
 		if err := udpServer.Shutdown(); err != nil {
-			log.Errorf("Shutting down UDP server failed: %v", err)
+			logrus.Errorf("Shutting down UDP server failed: %v", err)
 		}
 	}()
 
@@ -69,7 +70,7 @@ func StartVsockHost(ipv6 bool, hosts map[string]string, upstreamServers []string
 	return nil
 }
 
-func startDNSServer(vmGUID hvsock.GUID, vsockPort uint32, opts *rddns.ServerOptions) (*dns.Server, error) {
+func startDNSServer(vmGUID hvsock.GUID, vsockPort uint32, opts rddns.HandlerOptions) (*dns.Server, error) {
 	listner, err := vmsock.Listen(vmGUID, vsockPort)
 	if err != nil {
 		return nil, err
--- a/pkg/commands/start_host_windows.go
+++ b/pkg/commands/start_host_windows.go
@@ -14,7 +14,9 @@ limitations under the License.
 package commands
 
 import (
-	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
+	"github.com/miekg/dns"
+	rddns "github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
 	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/vmsock"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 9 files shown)


pkg/commands/start_host_windows.go
```diff
@@ -14,7 +14,9 @@ limitations under the License.
 package commands
 
 import (
-	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
+	"github.com/miekg/dns"
+	rddns "github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/dns"
 	"github.com/rancher-sandbox/rancher-desktop-host-resolver/pkg/vmsock"
```
mook-as, line 22: Shouldn't the RD imports go below the external imports?

pkg/dns/dns.go
```diff
@@ -25,54 +25,27 @@ type ServerOptions struct {
...
-	s := ``
-	for _, ip := range ips {
-		s += fmt.Sprintf("nameserver %s\n", ip)
-	}
-	r := strings.NewReader(s)
-	return dns.ClientConfigFromReader(r)
-}
-
-func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []string) (dns.Handler, error) {
+func NewHandler(opts *ServerOptions) (dns.Handler, error) {
```
mook-as: Shouldn't this be taking `HandlerOptions` rather than `ServerOptions`? If desired, `ServerOptions` can embed `HandlerOptions`.

Given that this is taking a pointer (instead of by value), shouldn't we check that `opts` isn't `nil`? (I'd just take a value instead though.)

I know upstream doesn't do it, but it would be good to have a doc comment on all exported things.

pkg/dns/dns.go
```diff
@@ -91,7 +64,7 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
 			}
 		}
 	} else {
-		if cc, err = newStaticClientConfig(upstreamServers); err != nil {
+		if cc, err = newStaticClientConfig(opts.UpstreamServers); err != nil {
```
mook-as, line 74: Silent fallback isn't great, but I guess we can leave it along for now (and see what upstream thinks).

pkg/dns/dns.go
```diff
@@ -103,29 +76,105 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
...
+
+func Cname(host string) string {
+	host = strings.ToLower(host)
+	if !strings.HasSuffix(host, ".") {
+		host += "."
+	}
+	return host
+}
+
+func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
```
mook-as, line 169: We no longer parse the incoming IPs before writing them to the file; that may be an issue for typos. It might still be useful to parse them at some point.

pkg/dns/dns.go
```diff
@@ -135,54 +184,24 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
 		}
 		switch q.Qtype {
 		case dns.TypeAAAA:
-			if !h.IPv6 {
+			if !h.ipv6 {
 				handled = true
```
mook-as: Upstream has some changes here (there's a big comment block and not setting `handle = true`); please pull that in.

pkg/dns/dns.go
```diff
@@ -135,54 +184,24 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
-			if _, ok := h.ip[cname]; ok {
-				addrs = []net.IP{h.ip[cname]}
-				err = nil
+			if _, ok := h.hostToIP[q.Name]; ok {
+				addrs = []net.IP{h.hostToIP[q.Name]}
 			} else {
-				addrs, err = net.LookupIP(cname)
+				addrs, err = net.LookupIP(q.Name)
+				if err != nil {
+					logrus.Errorf("handleQuery lookup IP failed: %v", err)
```
mook-as: This may be too much logging — this is just failing to lookup based on user input, which can happen all the time.

pkg/dns/dns.go
```diff
@@ -205,9 +224,35 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
 					handled = true
 				}
 			}
+		case dns.TypeCNAME:
+			cname := q.Name
+			if _, ok := h.cnameToHost[cname]; ok {
+				cname = h.cnameToHost[cname]
```
mook-as: This used to look up chains of cnames (`A` → `B` → `C` → `D` etc.); any reason to drop that?

pkg/dns/dns.go
```diff
@@ -205,9 +224,35 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
+			}
+			var err error
+			if _, ok := h.hostToIP[cname]; !ok {
+				cname, err = net.LookupCNAME(cname)
+				if err != nil {
+					logrus.Errorf("handleQuery lookup CNAME failed: %v", err)
```
mook-as: Probably shouldn't log as an error on user input here either.

pkg/dns/dns.go
```diff
@@ -205,9 +224,35 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
+				hdr.Rrtype = dns.TypeCNAME
+				a := &dns.CNAME{
+					Hdr:    hdr,
+					Target: cname,
+				}
+				reply.Answer = append(reply.Answer, a)
+				handled = true
+			}
 		case dns.TypeTXT:
 			txt, err := net.LookupTXT(q.Name)
```
mook-as, line 274: See upstream — there's something about breaking the answer up into 255-byte chunks again.

pkg/dns/dns.go
```diff
@@ -217,7 +262,11 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
 			}
 		case dns.TypeNS:
 			ns, err := net.LookupNS(q.Name)
-			if err == nil && len(ns) > 0 {
+			if err != nil {
+				logrus.Errorf("handleQuery lookup NS failed: %v", err)
+				continue
+			}
+			if len(ns) > 0 {
```
mook-as: FWIW, this is a pointless check: lookup over a list zero times is fine.

pkg/dns/dns.go
```diff
@@ -246,108 +299,64 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
-}
-
-func (h *Handler) ServeDNS(w dns.ResponseWriter, req *dns.Msg) {
-	switch req.Opcode {
-	case dns.OpcodeQuery:
-		h.handleQuery(w, req)
-	default:
-		h.handleDefault(w, req)
+	if err := w.WriteMsg(&reply); err != nil {
+		logrus.Errorf("handleDefault failed writing DNS reply: %v", err)
```
mook-as: We shouldn't log (at `Error`) about network issues in a long-running server.

pkg/dns/dns.go
```diff
@@ -103,29 +76,105 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
...
+		h.handleQuery(w, req)
+	default:
+		h.handleDefault(w, req)
+	}
+}
+
+func Start(opts *ServerOptions) (*Server, error) {
+	server := &Server{}
+	if opts.UDPPort > 0 {
+		udpOptions := opts
```
mook-as: Since `opts` is a pointer, this just makes a new pointer (and not a copy of the struct).  You never notice this because we clobber `TruncateReply` on the next call, and then never look at the struct again anyway.

pkg/dns/dns.go
```diff
@@ -103,29 +76,105 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
...
+		// always enable reply truncate for UDP
+		udpOptions.TruncateReply = true
+		h, err := NewHandler(udpOptions)
+		if err != nil {
+			return nil, err
+		}
+		addr := fmt.Sprintf("%s:%d", opts.Address, opts.UDPPort)
```
mook-as, line 122: I wonder about using [`net.JoinHostPort`](https://pkg.go.dev/net#JoinHostPort), but that takes a string rather than an int for the port… I guess not.

pkg/commands/start_host_windows.go
```diff
@@ -27,42 +29,60 @@ func StartVsockHost(ipv6 bool, hosts map[string]string, upstreamServers []string
...
-			IPv6:            ipv6,
-			StaticHosts:     hosts,
-			UpstreamServers: upstreamServers,
-			Listener:        streamListener,
-			TruncateReply:   false,
-		})
+	tcpServer, err := startDNSServer(vmGUID, vmsock.HostTCPListenPort, tcpOpts)
 	if err != nil {
-		return err
+		log.Panicf("StartVsockHost failed starting a TCP server: %v", err)
```
mook-as: Hmm, having a more user-understandable error message may be better. `failed to start TCP DNS server: %s`, maybe?

test/cmd/dnshammer.go
```diff
@@ -121,6 +126,22 @@ func compare(a, b []string) bool {
 	return true
 }
 
+func lookupCNAMERecord(domain string) (out []string, err error) {
+	// TODO (Nino-K): Switch back to the net.LookupCNAME once this is addressed: https://github.com/golang/go/issues/50101
+	c := dns.Client{}
+	m := dns.Msg{}
+	m.SetQuestion(domain, dns.TypeCNAME)
+	r, _, err := c.Exchange(&m, "127.0.0.1:53")
```
mook-as, line 120: Maybe we need a follow up to put `addr` back in…

test/cmd/dnshammer.go
```diff
@@ -121,6 +126,22 @@ func compare(a, b []string) bool {
...
+	if err != nil {
+		return nil, err
+	}
+	for _, a := range r.Answer {
+		cname := a.(*dns.CNAME)
+		out = append(out, cname.Target)
+	}
+	return out, err
```
mook-as, line 128: `return` because they're named, or (IMHO better) drop the names, declare `out` explicitly, and return `out, nil`.

test/e2e/dnslookup_test.go
```diff
@@ -331,11 +355,20 @@ func writeDNSHammerFile(path string, records map[string][]string) error {
 	return csvWriter.WriteAll(data)
 }
 
-func generateTXTrecords(n int) map[string][]string {
+func generateCNAMErecords(n int, domain string) map[string][]string {
+	records := make(map[string][]string)
+	for i := 1; i <= n; i++ {
+		subDomain := strings.ToLower(randomTxt(rand.Intn(10-1) + 1)) // nolint: gosec
```
mook-as: Presumably because of the rand? Would be good to actually comment on why it's disabled (as in, add a comment in the code).

test/e2e/dnslookup_test.go
```diff
@@ -331,11 +355,20 @@ func writeDNSHammerFile(path string, records map[string][]string) error {
...
+	}
+	return records
+}
+
+func generateTXTrecords(n int, domain string) map[string][]string {
 	records := make(map[string][]string)
-	baseDomain := "host-resolver-e2e-test"
 	for i := 1; i <= n; i++ {
-		records[fmt.Sprintf("%s%d.test.", baseDomain, i)] = generateTXT(rand.Intn(5-1) + 1) // nolint: gosec
+		records[fmt.Sprintf("%s-%d.test.", baseDomain, i)] = generateTXT(rand.Intn(5-1) + 1) // nolint: gosec
```
mook-as: Same as above — try to describe why we're disabling lint.

(It might be worth just mass disabling `gosec` in `*_test.go` via `.golangci.yaml`.)

mook-as (changes requested):
(I'm reviewing this by diffing against upstream `dns.go` instead.)

pkg/dns/dns.go
Nino-K: > Shouldn't this be taking HandlerOptions rather than ServerOptions? If desired, ServerOptions can embed HandlerOptions.

I think we should just rename `ServerOptions` --> `HandlerOptions`

> I know upstream doesn't do it, but it would be good to have a doc comment on all exported things.
They are not really following the exported function/comment Scheme, I think it's best to leave it since it aligns with the rest of the project.

jandubois: > They are not really following the exported function/comment Scheme, I think it's best to leave it since it aligns with the rest of the project.

I think adding the comments will be fine, especially if it is in a part of the code that is re-used outside the project.

test/e2e/dnslookup_test.go
Nino-K: I will do a separate PR for this as a follow-up PR.

pkg/dns/dns.go
```diff
@@ -171,183 +267,128 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
+				continue
+			}
+			if h.truncate {
+				logrus.Debugf("handleDefault truncating reply: %v", reply)
 				reply.Truncate(truncateSize)
-				_ = w.WriteMsg(reply)
-				return
 			}
+			if err = w.WriteMsg(reply); err != nil {
+				logrus.Errorf("handleDefault failed writing DNS reply to [%v]: %v", addr, err)
```
mook-as: Missed one error

pkg/dns/dns.go
```diff
@@ -103,63 +83,182 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
...
+	}
+}
+
+func Start(opts ServerOptions) (*Server, error) {
+	server := &Server{}
+	if opts.UDPPort > 0 {
+		udpOptions := opts
+		// always enable reply truncate for UDP
+		udpOptions.TruncateReply = true
+		h, err := NewHandler(opts.HandlerOptions)
```
mook-as: Shouldn't this be using `udpOptions`?

pkg/dns/dns.go
```diff
@@ -103,63 +83,182 @@ func newHandler(truncate, IPv6 bool, hosts map[string]string, upstreamServers []
...
+		server.udp = s
+		go func() {
+			logrus.Debugf("Start UDP server listening on: %v", addr)
+			if e := s.ListenAndServe(); e != nil {
+				panic(e)
+			}
+		}()
+	}
+	if opts.TCPPort > 0 {
+		tcpOpts := opts
```
mook-as, line 133: This should mirror `udpOptions` (opts vs options).

mook-as (changes requested):

pkg/dns/dns.go
Nino-K: 🤔 I totally missed it after the factoring.

mook-as (approved):

PR #28, opened 2022-07-16
Don't display cobra usage after failure

Nino-K:


ericpromislow (approved):
Verified that if we give an invalid option to a subcommand, we get usage for that command. But not after.

PR #30, opened 2022-07-20
Use the actual function argument

Nino-K:


mook-as (approved):
I don't understand how things worked.

PR #33, opened 2022-10-17
Bumps lima-vm dns

Nino-K:


jandubois (approved):
Thanks, LGTM.

As discussed in person, this PR includes the effects of `go mod tidy`, which is required to pass linting with the current Go version. The alternative would be to upgrade Go.

Since this only upgrades `cobra` and `mousetrap` in addition to bumping Lima, this seems acceptable.

PR #34, opened 2022-10-19
Fixes e2e to not assume WSL 2

Nino-K:

### DIFF
--- a/test/e2e/dnslookup_test.go
+++ b/test/e2e/dnslookup_test.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/csv"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -44,9 +45,13 @@ import (
 
 var (
 	// for now we use this maybe this can be an env var from test
-	wslDistroName         = "host-resolver-e2e-test"
-	wslTarballName        = "distro-0.21.tar"
-	wslTarballURL         = "https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.21/distro-0.21.tar"
+	wslDistroName  = "host-resolver-e2e-test"
+	tarbalVersion  = "0.27"
+	wslTarballName = fmt.Sprintf("distro-%s.tar", tarbalVersion)
+	wslTarballURL  = fmt.Sprintf(
+		"https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v%s/%s",
+		tarbalVersion,
+		wslTarballName)
 	testSrvAddr           = "127.0.0.1"
 	dnsPort               = "53"
 	dnsHammerArecords     = "testA.csv"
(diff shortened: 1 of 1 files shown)


test/e2e/dnslookup_test.go
```diff
@@ -130,20 +133,32 @@ func TestMain(m *testing.M) {
 	err = writeDNSHammerFile(filepath.Join(tmpDir, dnsHammerCNAMERecords), cnameRecords)
 	requireNoErrorf(err, "Failed generating CNAME test data")
 
-	logrus.Infof("Dowloading %v wsl distro tarball", wslTarballName)
-	tarballPath := filepath.Join(tmpDir, wslTarballName)
-
-	err = downloadFile(tarballPath, wslTarballURL)
-	requireNoErrorf(err, "Failed to download wsl distro tarball: %v", err)
+	tarballPath := filepath.Join(os.TempDir(), wslTarballName)
```
jandubois, line 140: Why don't you use `tmpdir` here?

```suggestion
	tarballPath := filepath.Join(tmpdir, wslTarballName)
```

test/e2e/dnslookup_test.go
```diff
@@ -130,20 +133,32 @@ func TestMain(m *testing.M) {
...
+	s, err := os.Stat(tarballPath)
+	// Only download the tarball if
+	// 1) it doesn't exist
+	// 2) the cache is expired > 60 min
```
jandubois: Why do you care about this? The file should never change, so why would it expire?

test/e2e/dnslookup_test.go
```diff
@@ -130,20 +133,32 @@ func TestMain(m *testing.M) {
...
+	if errors.Is(err, os.ErrNotExist) {
+		err = downloadFile(tarballPath, wslTarballURL)
+		requireNoErrorf(err, "Failed to download wsl distro tarball %v", wslTarballName)
+	} else {
+		if s.ModTime().Before(time.Now().Add(time.Hour * -1)) {
```
mook-as: If `err` is non-`nil` (and not `ErrNotExist`) but we haven't timed out yet, this drops the error on the floor.

mook-as (changes requested):

test/e2e/dnslookup_test.go
Nino-K: This is intentional, `tmpdir` is only available within the life of each test run and it gets removed after each run, however,  `os.TmpDir()` is bound to the user's OS session, which makes it more ideal location for caching.

test/e2e/dnslookup_test.go
Nino-K: We don't necessarily care about the changes in the file, this is just a simple caching mechanism to not download the file each time. During development, this saves tons of time since it reduces the test time by 80%.

test/e2e/dnslookup_test.go
jandubois: You are literally setting `tmpDir = os.TempDir()` just 20 lines earlier; how can this be a different value now?

jandubois: Just noticed I wrote `tmpdir` instead of `tmpDir` in the suggestion above; I meant the latter. But there is no `tmpdir` in the code, so I think this was still clear.

Nino-K: I was under the assumption that host-resovler e2e is doing the same thing as Vtunnel [here](https://github.com/rancher-sandbox/rancher-desktop/blob/1e31560b06c62da5333d7c4f5e75ea86d1915ba3/src/go/vtunnel/test/e2e/connectivity_test.go#L255) since I was working on them in parallel, I will update.

mook-as: From discussing with Nino offline: this is intentional so that we _don't_ delete the tarball (so it can be reused on next run).

mook-as (approved):

